PR #27, opened 2021-03-25
Support for ROOT version 6.24 (RC1)

Vindaar:
I built a local ROOT today for RC1 of version 6.24.

First of all it doesn't build on my machine unless I build with C++14 support. I *think* the (while old) GCC version on sultan already supports a big subset (or possibly the full) C++14 standard. In that case this shouldn't cause any issues.

After changing this, GCC complained about the `!=` comparison in `TRestBrowser.cxx`. Turns out all of a sudden it cannot resolve the correct overload to choose anymore. Adding the manual type conversion to one of the overloads makes it work.

Feel free to merge at a later time if C++14 seems too *futuristic* at the moment.


jgalan (approved):
It is fine with me.

nkx111 (commented):
We still need c++11 in SJTU. What if we remain CMakeLists.txt unchanged and only update TRestBrowser.cxx? Will the compilation fail?

lobis:
I am also a supporter of upgrading to C++14, I think its time.

nkx111 (approved):

Vindaar:
@nkx111 
> We still need c++11 in SJTU. What if we remain CMakeLists.txt unchanged and only update TRestBrowser.cxx? Will the compilation fail?

What changed your mind to approve the PR?

And no, with C++11 I get all sorts of other compilation errors.

nkx111:
@Vindaar We are using REST v2.2 here in SJTU so there is no conflict actually. I don't mind upgrading to c++14 on v2.3.

PR #33, opened 2021-04-07
New logos updated

jgalan:
Improved logo and uploaded source powerpoint files.


lobis (commented):
I think we should have the main logo in SVG format.

jgalan:
> I think we should have the main logo in SVG format.

Ok, just add it

lobis:
> > I think we should have the main logo in SVG format.
> 
> Ok, just add it

Its not possible to convert png to svg since its a vector format, but it is possible to convert svg to png. Vector formats allow for unlimited zoom as they consist of geometrical shapes instead of pixels and are the best format for logos, since all the other pixelized formats can be derived from it.

We would need to redraw the logo using a program like Inkscape, I could do it if you want, but it will not look 100% identical to the original.

jgalan:
The original file is in PPT, cannot be transformed to SVG?

lobis:
> The original file is in PPT, cannot be transformed to SVG?

Good Idea. I will upload the logos in SVG.

lobis (changes requested):
Upload logos in SVG

lobis (approved):

PR #41, opened 2021-04-14
Development

jgalan:
This merge to master is just to update the visible items at the README.md when we are at master branch.

I believe it is good to keep pull-request to `development` branch, and do pull-request from `development` to `master`.

Even admins are forced to ask for review when doing a pull-request to master. Although admins are allowed to force the merge (at `development` branch) without review, it is always good to allow for review when pull-request to `development`.


lobis (approved):

.gitlab-ci.yml
```diff
@@ -1,4 +1,4 @@
-image: nkx1231/root6-geant4-garfield:0.6
+image: lobis/root-geant4-garfieldpp
```
nkx111, line 1: I still don't know why we have to switch the docker image

jgalan: The code was not compiling with your docker image ...

https://gitlab.cern.ch/rest-for-physics/framework/-/jobs/13162589

After updating the image it would compile again

https://gitlab.cern.ch/rest-for-physics/framework/-/commit/a489ac3e29ead42e1c86d21f8b166ad7453017bc/pipelines?ref=development

jgalan: Of course, I believe we could make it work with both images.

Doing a quick search I did not find how to run both images in gitlab in parallel. I.e. launching both images in two pipelines when we push to

nkx111: Is it because the previous updated c++14 compilation flag?

It seems not a big error to fix.

nkx111: > Doing a quick search I did not find how to run both images in gitlab in parallel.

Maybe that solve the problem

https://stackoverflow.com/questions/33387622/multiple-docker-images-in-gitlab-ci-yml

jgalan:
It might be handy to have this here [pipeline status](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/development)

jgalan:
Next problem arrives. We need to do not source `/../.../../../geant4.sh` when using `nxk-` image. Not sure if there is a `CI_IMAGE` variable?

jgalan:
> Next problem arrives. We need to do not source `/../.../../../geant4.sh` when using `nxk-` image. Not sure if there is a `CI_IMAGE` variable?

Ok, problem solved now. Perhaps inside the docker there is a variable that we can use to identify it?

jgalan:
Any clue why is not succeeding now?

nkx111:
> Any clue why is not succeeding now? 

It's too strange. I don't know why the last step of loading `thisREST.sh` failed with no error message. The file content seems normal. I have removed this command in `build_nkx` job. That makes it success.

nkx111 (approved):

jgalan:
It seems there is a limit on the number of times we may pull @lobis image! It is not the case for @nkx?

https://gitlab.cern.ch/rest-for-physics/framework/-/jobs/13258153

lobis:
> It seems there is a limit on the number of times we may pull @lobis image! It is not the case for @nkx?
> 
> https://gitlab.cern.ch/rest-for-physics/framework/-/jobs/13258153
> 
> It seems the limitation is 100pulls/6hours for anonymous users.
> 
> I noticed that at each stage is pulling the image. I wonder why this is needed at each stage, and if there is a way to avoid that.

I think this is a common restriction to all free users, since November of 2020. I don't think images created before this date are exempt (but it could be). Looking at the numbers I see that my image has ~600 pulls as of right now, while @nkx111's has ~10k (but his image has been up for quite more time) so I think since we didn't use his image in the lfna pipeline we never triggered the limit. I think the free limit is quite reasonable so there shouldn't be an issue once we optimize our pipelines a bit more.

There are other alternatives to dockerhub, like github itself, but I think its more restrictive.

PR #44, opened 2021-05-05
Refactoring

lobis:
The only functional change is the implementation via precompiler directive of a scaling in the visualization (ecb279b).

The refactoring consists on spell check, replacing NULL by nullptr (better practice) and organizing dependencies a bit better.


lobis:
I have updates the restG4 CMakeLists.txt file with the new `find_package` strategy (https://github.com/rest-for-physics/restG4/blob/f42b4b062ea6b49e10bf95781b33c6d98cecfc76/CMakeLists.txt).

I am calling `rest-config` from inside the `RESTConfig.cmake` which probably is not the best way to do this, but it should work just fine (as long as `rest-config` does not change). Maybe it would be nice to find the correct way to do this.

jgalan:
I believe `GEOM_SCALE` was introduced long time ago by @juanangp.

It is very interesting the `find_package` possibility.

Some time ago I though even about the possibility to introduce a function `rest_minimum_required(2,3,5)` that we could include at the libraries to force using a minimum framework version. To prevent someone at cmake level that a particular library pull will not compile if REST framework is not updated. This happened in the past, but it was solved by asking users to stick to the official version, although sometimes this might be not possible.

jgalan (approved):
Since pipelines are passing I see no problem to approve the changes. I let @nkx and @juanangp  to have a look too.

CMakeLists.txt
```diff
@@ -36,29 +35,29 @@ IF (NOT DEFINED RESTLIB_MODEL)
     set(RESTLIB_MODEL OFF)
 ENDIF ()
 
-#auto schema evolution for ROOT
+# Auto schema evolution for ROOT
```
jgalan, line 38: There is a parsing tool you are using to reformat this?

lobis: > There is a parsing tool you are using to reformat this?

No, I did it by hand while I read the code. I think it's more legible this way, don't you agree?

jgalan: Yes, looks more tidy

source/framework/core/inc/TRestEveEventViewer.h
```diff
@@ -36,10 +27,22 @@
...
+#include <TObject.h>
+#include <TPad.h>
+#include <TVector3.h>
 
+#include "TRestEventViewer.h"
+
+#if ROOT_VERSION_CODE >= ROOT_VERSION(6, 24, 0)
+#define GEOM_SCALE 1
+#else
 #define GEOM_SCALE 0.1
```
juanangp, line 44: I don't remember about this.

Looks like GEOM_SCALE is an scale factor to display the deposited energy. and it is just used here:
TRestEveEventViewer::AddSphericalHit(double x, double y, double z, double r
adius, double en) {

I would suggest to remove the define and either remove GEOM_SCALE or add an rml variable about that. Perhaps, not in this pull request.

source/framework/core/inc/TRestEvent.h
```diff
@@ -23,14 +23,14 @@
 #ifndef RestCore_TRestEvent
 #define RestCore_TRestEvent
 
-#include <iostream>
-
+#include <TBufferFile.h>
+#include <TObject.h>
 #include <TPad.h>
+#include <TRestTools.h>
```
juanangp, line 29: Change to #include "TRestTools.h", I assume this is in our repo.

source/framework/core/inc/TRestEventViewer.h
```diff
@@ -19,20 +19,19 @@
 #ifndef RestCore_TRestEventViewer
 #define RestCore_TRestEventViewer
 
+#include <TGButton.h>
+#include <TGLabel.h>
+#include <TGTextEntry.h>
 #include <TGeoManager.h>
 #include <TObject.h>
+#include <TRestEvent.h>
```
juanangp, line 27: Same #include "TRestEvent.h"

source/framework/core/inc/TRestEventViewer.h
```diff
@@ -19,20 +19,19 @@
...
-#include "TGButton.h"
-#include "TGLabel.h"
-#include "TGTextEntry.h"
-#include "TRestEvent.h"
-
 class TRestBrowser;
 class TRestEventViewer {
    protected:
-    TGeoManager* fGeometry = 0;     //!
+    TGeoManager* fGeometry = 0;  //!
```
juanangp, line 34: TGeoManager* fGeometry = nullptr;

source/framework/core/inc/TRestGDMLParser.h
```diff
@@ -2,9 +2,10 @@
 #ifndef restG4_TRestGDMLParser
 #define restG4_TRestGDMLParser
 
-#include "TGeoManager.h"
+#include <TGeoManager.h>
```
juanangp, line 5: #include "TGeoManager.h"? I am confused since typically you are doing the right thing. I am missing something? TGeoManager belongs to restsoft right?

lobis: TGeoManager is a ROOT class.

source/framework/core/inc/TRestEveEventViewer.h
lobis: Why would you add an RML variable? this is just a constant, it shouldn't be modified by the user. It appears because you need to rescale all positions by 0.1 in order for them to be in line with the displayed volume by `TGeoManager`. This has always seemed off to me, but its nothing related to REST (In any ROOT macro you also need to divide by 10).

I guess they caught up to this "bug" and fixed it in the newest versions, so you no longer need to do this, thats why I implemented this logic.

source/framework/core/inc/TRestEventViewer.h
lobis: I will remove this instead (`TGeoManager* fGeometry = 0` -> `TGeoManager* fGeometry;`) since if I am not wrong there is no reason to initialize them to nullptr besides readability right?

source/framework/core/inc/TRestEveEventViewer.h
juanangp: Well, to add an RML variable is not required. But there is no reason to have a #define inside a header which is only used in TRestEveEventViewer::AddSphericalHit(double x, double y, double z, double r
adius, double en)

Can you just move this hunk:
#if ROOT_VERSION_CODE >= ROOT_VERSION(6, 24, 0)
#define GEOM_SCALE 1
#else
#define GEOM_SCALE 0.1
inside TRestEveEventViewer::AddSphericalHit

source/framework/core/inc/TRestEventViewer.h
juanangp: Looking at source/framework/core/src/TRestEventViewer.cxx looks like you can call TRestEveEventViewer::SetGeometry(TGeoManager* geo) with fGeometry un-initialized, there is actually a protection about that. I would suggest to keep the initialization as nullptr, shouldn't be a problem.

jgalan: Allways better to initialize

PR #51, opened 2021-05-24
Iaxo data pipeline

KonradAltenmueller:
Added the pipeline for the iaxo data. As of now only the veto process is validated, but a raw data file and a complete analysis rml is added, so that it can be easily expanded for other processes. 
The veto validation checks if the correct signal IDs are removed from the raw events, and if the veto observables are created.

Pipeline validation badge: [pipeline status](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/iaxo_data_pipeline)


jgalan:
I believe validation is failing because you are using python3, but ROOT6 at @nkx docker is linking to python2?

nkx111:
I see, it is because libpython2.7 is used by root6.10. It was not installed and therefore pyROOT cannot be used. I have added this library in nkx1231/root6-geant4-garfield:0.8

jgalan:
I changed the image being used for validation but the problem persists

https://gitlab.cern.ch/rest-for-physics/framework/-/jobs/14030389

nkx111:
I just changed it to use the system default python. In nkx image it is python2.7. Now it works.

nkx111 (approved):

jgalan (commented):
Hi @KonradAltenmueller I just believe that this validation job could have been placed directly at the `rawlib` repository. Together with addNoise.

I guess you do not need the readout and so on. And all detector processes are disabled at the RML.

That would simplify the pipeline, and placed dedicated rawlib tests fully at an independent library.

You are still on time to avoid pushing these changes to the framework, and create a pull-request at rawlib, uploading just what is needed.

A main framework validation will become necessary when you need to validate a long data processing chain that uses different libraries.

KonradAltenmueller:
Yes, I can move it to `rawlib`.
But don't you think it makes sense to have it in the framework pipeline, and add validations for the other processes in the iaxo-analysis chain? In a similar way as for trex for example...

PR #72, opened 2021-06-18
Add `TRestExpressionEvaluationProcess`

Vindaar:
Large Vindaar  [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/addTRestExpressionEvaluationProcess) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=rest-for-physics&utm_campaign=badge_info)

I finally found the time and motivation to go back to this and finish it.

I've added support for:
- map access: `foo[5]`
- also allow filtering instead of only mapping

Those items listed down in this example as `filter` act, well, as a filter. That means if the boolean expression is true, the event is kept, else it's dropped from the output tree.

```xml
///   <expressionset>
///     <!-- a simple, constant boolean expression -->
///     <item name="boolExpr" expr="5<10"/>
///     <!-- a simple, constant float expression -->
///     <item name="floatExpr" expr="5 * 10"/>
///     <!-- a float expression using an existing REST observable -->
///     <Item name="floatExprObs" expr="hitsAna_energy / 1000.0"/>
///     <!-- a bool expression using an existing REST observable -->
///     <item name="boolExprObs" expr="hitsAna_energy < 5000.0"/>
///     <!-- a bool expression of multiple statements combined by an `and` -->
///     <item name="boolAndExprMultiple" expr="hitsAna_energy < 5000.0 and tckAna_nTracks_X==1"/>
///     <!-- a bool expression of multiple statements combined by an `or` -->
///     <item name="boolOrExprMultiple" expr="hitsAna_energy < 5000.0 or tckAna_nTracks_X==1"/>
///     <!-- a complicated expression of nested parenthesis and multiple expressions -->
///     <item name="boolComplicated" expr="(hitsAna_energy / 1000.0) < 5.0 or (tckAna_nTracks_X==1 and tckAna_nTracks_Y==1)"/>
///     <!-- an expression used as a filter. I.e. if the predicate is true we keep the event, else drop it -->
///     <item filter="hitsAna_energy < 3000"/>
///     <!-- a mapping expression accessing a specific element from an observable that is map<int, Double> -->
///     <item name="veto4638" expr="peakTime[4638]"/>
///     <!-- a filtering expression accessing an observable that is map<int, Double> -->
///     <item filter="peakTime[4638] < 300"/>
///   </expressionset>
```

However, I can't quite figure out the best way to add something to the CI. It's too opaque for me. What's the typical way to add ROOT files to use as input for the CI? 
I don't know:
- where the CI is run (is it a local machine with access to sultan? or is it on public Gitlab resources?)
- I don't have the files locally that are used, e.g. those 01.NLDBD files etc, so I don't know if I can reuse some of them.

Personally, I'd like to just start from a fully reconstructed ROOT tree (small one with ~1000 events or so) and apply different RML files and check if the outputs match my expectation. Please give me some guidance on how to make such a file available in the CI. Maybe there's a way to host it on sultan, same as that trex readout file that's being `wget`ed?

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -361,7 +361,23 @@ trexDM_data_latest:
     variables:
       - $CRONJOB == "YES"
 
-pandaXIII_Topological: 
+expressionEvaluation:
+  type: restManager_process
+  script:
+    - . ${CI_PROJECT_DIR}/install/thisREST.sh
+    - cd ${CI_PROJECT_DIR}/pipeline/expressionEvaluation
+    - cp ${CI_PROJECT_DIR}/pipeline/trex/Hits_01929.root .
+    - restManager --c expressionEvalCi.rml --f Hits_01929.root
+    - root -b -q validateExpressionEval.C
+  artifacts:
+    paths:
+      - ${CI_PROJECT_DIR}/pipeline/expressionEvaluation/Hits_exprEval_01929.root
+    expire_in: 1 day
+  except:
+      variables:
+        - $CRONJOB == "YES"
+
+pandaXIII_Topological:
   type: restManager_process
   script:
     - . ${CI_PROJECT_DIR}/install_latest/thisREST.sh
--- a/pipeline/expressionEvaluation/expressionEvalCi.rml
+++ b/pipeline/expressionEvaluation/expressionEvalCi.rml
@@ -8,36 +8,36 @@
 </globals>
 
 <TRestRun name="ExpressionEvalCI" title="ExpressionEvalCI" verboseLevel="silent">
-	<parameter name="experiment" value="IAXOD0"/>
-	<parameter name="runNumber" value="preserve"/>
-	<parameter name="runTag" value="preserve"/>
-	<parameter name="runType" value="${RUN_TYPE}"/>
-	<parameter name="runDescription" value=""/>
-	<parameter name="user" value="${USER}"/>
-	<parameter name="verboseLevel" value="0"/>
-	<parameter name="overwrite" value="off" />
+        <parameter name="experiment" value="IAXOD0"/>
+        <parameter name="runNumber" value="preserve"/>
+        <parameter name="runTag" value="preserve"/>
+        <parameter name="runType" value="${RUN_TYPE}"/>
+        <parameter name="runDescription" value=""/>
+        <parameter name="user" value="${USER}"/>
+        <parameter name="verboseLevel" value="0"/>
+        <parameter name="overwrite" value="off" />
 
-	<parameter name="inputFormat" value="Hits_[fRunNumber].root"/>
-	<parameter name="outputFile" value="Hits_exprEval_[fRunNumber].root" />
-	<parameter name="readOnly" value="false" />
+        <parameter name="inputFormat" value="Hits_[fRunNumber].root"/>
+        <parameter name="outputFile" value="Hits_exprEval_[fRunNumber].root" />
+        <parameter name="readOnly" value="false" />
 
 </TRestRun>
-	<TRestProcessRunner name="EvalSignals" title="EvalSignals" > <!-- <verboseLevel="info"> -->
-		<parameter name="firstEntry" value="0" />
-		<parameter name="lastEntry" value="0" />
-		<parameter name="eventsToProcess" value="0" />
-		<addProcess type="TRestExpressionEvaluationProcess" name="Testing" value="ON" verboseLevel="silent">
+        <TRestProcessRunner name="EvalSignals" title="EvalSignals" > <!-- <verboseLevel="info"> -->
+                <parameter name="firstEntry" value="0" />
+                <parameter name="lastEntry" value="0" />
+                <parameter name="eventsToProcess" value="0" />
+                <addProcess type="TRestExpressionEvaluationProcess" name="Testing" value="ON" verboseLevel="silent">
                   <expressionset>
-		    <item name="testExpr" expr="5<10"/>
+                    <item name="testExpr" expr="5<10"/>
                     <item filter="rawAna_risetime_map[692] < 25"/>
                     <item name="rawAna_test" expr="rawAna_risetime_map[692] < 20"/>
-		    <item filter="hitsAna_energy < 5000"/>
-		    <item name="someBool" expr="hitsAna_energy < 5000 and hitsAna_xySigmaBalance < 0.0"/>
+                    <item filter="hitsAna_energy < 5000"/>
+                    <item name="someBool" expr="hitsAna_energy < 5000 and hitsAna_xySigmaBalance < 0.0"/>
                     <item name="rawAnaRisetime692" expr="rawAna_risetime_map[692]"/>
-		    <item name="milliEnergy" expr="hitsAna_energy / 1000.0"/>
+                    <item name="milliEnergy" expr="hitsAna_energy / 1000.0"/>
                   </expressionset>
-		</addProcess>
+                </addProcess>
 
-	</TRestProcessRunner>
-	<addTask type="processEvents" value="ON" />
+        </TRestProcessRunner>
+        <addTask type="processEvents" value="ON" />
 </TRestManager>
--- a/pipeline/expressionEvaluation/expressionEvalCi.rml
+++ b/pipeline/expressionEvaluation/expressionEvalCi.rml
@@ -1,0 +1,43 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no" ?>
+
+<TRestManager>
+
+<globals>
+    <parameter name="mainDataPath" value="." />
+    <variable name="RUN_TYPE" value="RawToSignal" overwrite="false" />
+</globals>
+
+<TRestRun name="ExpressionEvalCI" title="ExpressionEvalCI" verboseLevel="silent">
+	<parameter name="experiment" value="IAXOD0"/>
+	<parameter name="runNumber" value="preserve"/>
+	<parameter name="runTag" value="preserve"/>
+	<parameter name="runType" value="${RUN_TYPE}"/>
+	<parameter name="runDescription" value=""/>
+	<parameter name="user" value="${USER}"/>
+	<parameter name="verboseLevel" value="0"/>
+	<parameter name="overwrite" value="off" />
+
+	<parameter name="inputFormat" value="Hits_[fRunNumber].root"/>
+	<parameter name="outputFile" value="Hits_exprEval_[fRunNumber].root" />
+	<parameter name="readOnly" value="false" />
+
+</TRestRun>
+	<TRestProcessRunner name="EvalSignals" title="EvalSignals" > <!-- <verboseLevel="info"> -->
+		<parameter name="firstEntry" value="0" />
+		<parameter name="lastEntry" value="0" />
+		<parameter name="eventsToProcess" value="0" />
+		<addProcess type="TRestExpressionEvaluationProcess" name="Testing" value="ON" verboseLevel="silent">
+                  <expressionset>
+		    <item name="testExpr" expr="5<10"/>
+                    <item filter="rawAna_risetime_map[692] < 25"/>
+                    <item name="rawAna_test" expr="rawAna_risetime_map[692] < 20"/>
+		    <item filter="hitsAna_energy < 5000"/>
+		    <item name="someBool" expr="hitsAna_energy < 5000 and hitsAna_xySigmaBalance < 0.0"/>
+                    <item name="rawAnaRisetime692" expr="rawAna_risetime_map[692]"/>
+		    <item name="milliEnergy" expr="hitsAna_energy / 1000.0"/>
+                  </expressionset>
+		</addProcess>
+
+	</TRestProcessRunner>
+	<addTask type="processEvents" value="ON" />
+</TRestManager>
--- a/pipeline/expressionEvaluation/validateExpressionEval.C
+++ b/pipeline/expressionEvaluation/validateExpressionEval.C
@@ -26,17 +26,17 @@
     map<int, int> rawAnaHits692Vals;
     // walk through tree
     for(int i = 0; i < t->GetEntries(); i++){
-	t->GetEntry(i);
-	// break if milli energy does not match real energy
-	if((milliEnergy * 1000.0 - hitsAna_energy) > 1e-6) return 2;
-	// testExpr is just 5<10, so always true
-	if(!testExpr) return 3;
-	if(someBool) someBoolCountTrue++;
-	// have a filter for energy to be < 5000
-	if(hitsAna_energy >= 5000) return 4;
-	if(rawAna_test) rawAnaTestCountTrue++;
-	// increase count
-	rawAnaHits692Vals[rawAnaRisetime692]++;
+        t->GetEntry(i);
+        // break if milli energy does not match real energy
+        if((milliEnergy * 1000.0 - hitsAna_energy) > 1e-6) return 2;
+        // testExpr is just 5<10, so always true
+        if(!testExpr) return 3;
+        if(someBool) someBoolCountTrue++;
+        // have a filter for energy to be < 5000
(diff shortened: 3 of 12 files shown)


jgalan:
Hi, there is a problem, you should pull-request from a branch you created at the framework so that validation tests run at gitlab.cern.ch

jgalan:
The pipeline is executing here:

https://gitlab.cern.ch/rest-for-physics/framework/

At the file `.gitlab-ci.yml`, If you place your validation test after another stage generating an *artifact*, then you will be able to use that artifact. I.e. at your validation routine you will be able to use the following file:

```
        - root -b -q ../ValidateTrees.C'("validation.root")'    -
        -  restRoot -b -q ValidateDetectorParams.C'("Hits_01929.root")'  
    artifacts:    
        paths:      
        - ${CI_PROJECT_DIR}/pipeline/trex/Hits_01929.root    
        expire_in: 1 day
```

Here you could use the file generated by trex-dm validation pipeline, `Hits_01929.root`, for example.

Vindaar:
> Hi, there is a problem, you should pull-request from a branch you created at the rest-for-physics/framework so that validation tests run at gitlab.cern.ch

I might do that, but I don't have write permission to this repository. ;)
Would be handy if it worked from forks as well though.

> At the file .gitlab-ci.yml, If you place your validation test after another stage generating an artifact, then you will be able to use that artifact. I.e. at your validation routine you will be able to use the following file:

Thanks, but I was hoping there'd be an easy way to start from a file that's a bit more suited to the testing. I haven't checked out the trex file though. I'll download that artifact and look at the file.

Vindaar:
Ok, I've added some tests now. If you give me access to the repository I can recreate the PR from the rest repo.

jgalan:
> I might do that, but I don't have write permission to this repository. ;)

Ok, you need to be a member of `rest-for-physics` organization. I believe GitHub applied changes recently, so that organizations manage repository access using *teams*. You should accept the invitation.

> Would be handy if it worked from forks as well though.

Not sure how to do that, we use gitlab at CERN to mirror few repositories. It was be configured manually.

If you have a CERN account and login into gitlab.cern.ch, I will be able to add you as a member at `rest-for-physics` at CERN, so that you are allowed to force an update. If not, I believe updates haven every half-an-hour.

> Thanks, but I was hoping there'd be an easy way to start from a file that's a bit more suited to the testing. I haven't checked out the trex file though. I'll download that artifact and look at the file.

Right, making a simple test would be more elegant. But connecting it to a complex chain will  help to validate other inter-dependencies. I suggest you launch a processing chain with just your new process, operating over `hitsAna` observables.

jgalan:
> Those items listed down in this example as filter act, well, as a filter. That means if the boolean expression is true, the event is kept, else it's dropped from the output tree.

I believe any REST process is able to do that using the `<cut variable="sAna_BaseLineSigmaMean" condition=">12" value="ON" />`

I believe the parameter `cutsEnabled` should be true.

We should make sure this is not replicated, and it would be good we all use the same standards. @nkx?

jgalan:
Another 2 comments I have:

   1. why you need to define an `<expressionset>` ? I would say your process, c++ object, is an expression set. You may connect any number of processes you wish in your data chain, and each "expressionset" connected with each process, will be identified by the process name.
   2. So, it is important that we follow the same naming conventions for observables, i.e. `pcsName_obsName`. Why not calling directly `SetObservableValue((string)exprPair.first, value.getLeft());` instead of `fAnalysisTree->SetObservableValue((string)exprPair.first, value.getLeft());`

jgalan:
> * where the CI is run (is it a local machine with access to sultan? or is it on public Gitlab resources?)

At each README.md where we have implemented a pipeline there is a badge that allows you to click and you will be lead directly to the last pipeline results.

As this badge.
[pipeline status](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/master)

Vindaar:
> Right, making a simple test would be more elegant. But connecting it to a complex chain will help to validate other inter-dependencies. I suggest you launch a processing chain with just your new process, using HitsXX.root as input, and operating over hitsAna observables.

Good point as well.

> I believe any REST process is able to do that using the <cut variable="sAna_BaseLineSigmaMean" condition=">12" value="ON" />
>
> Although perhaps it would be simpler to write <cut condition="sAna_BaseLineSigmaMean > 12" />.
>
> I believe the parameter cutsEnabled should be true.
>
> We should make sure this is not replicated, and it would be good we all use the same standards. @nkx?

I agree that duplication isn't a good thing. However, the supported expressions here are more powerful than those supported by `<cut *`:
- arbitrary nesting of multiple boolean & double expressions with or without parentheses (used for precedence of course)
- support for math operations (`*`, `/`, `+`, `-`) and more boolean operations (unary `not`). Can be easily extended to support more, e.g. `xor` or math functions `sqrt` etc.
- supports bracket expressions to access individual fields of a map
- string is parsed *once* into a tree that is evaluated for each event. In contrast to parsing the string for each event, making it much more efficient.

In addition the name `cut` is confusing, if things for which the predicate is true will be kept (those should be dropped if we have a "cut"). `filter` is the more natural name in functional programming for such an operation.

Finally, having a single process that is used to handle such a feature makes it clearer to use than having random "cuts" all over the place.

Given this feature disparity, I don't think it makes sense to consider this duplication. But yes, this makes the existing `cut` feature obsolete.

> why you need to define an <expressionset> ? I would say your process, c++ object, is an expression set. > You may connect any number of processes you wish in your data chain, and each "expressionset" will be > connected with each TRestExpressionEvaluationProcess, and it will be identified by the process name.

For three reasons mainly:
1. it makes it clearer to the user that the following is a kind of domain specific language that starts and stops clearly.
2. it's easier to implement if I can start cleanly from a node `expressionset` where I control the things that are valid inside, instead of parsing all fields of the process. 
3. it allows to add functionality outside the `expressionset` at some point if so desired. For instance defining variables that can be used inside of an expression in case one wishes to apply environmental variables to expressions. In that case it might be cleaner to do define those variables once outside and then use the defined name inside of all expressions than having to possibly write `${myEnvVariable}` in multiple expressions.

To be honest, it's mainly preference due to 1 and 2.

> So, it is important that we follow the same naming conventions for observables, i.e. pcsName_obsName. Why not calling directly SetObservableValue((string)exprPair.first, value.getLeft()); instead of fAnalysisTree->SetObservableValue((string)exprPair.first, value.getLeft());. So that as soon as I do run0->PrintObservables I know where the observables came from.

I don't have any preferences here. If you wish I can make that change. Without having checked the code, how is the prefix name determined?

jgalan:
> I don't have any preferences here. If you wish I can make that change. Without having checked the code, how is the prefix name determined?

The prefix is the name of the process. So if you define your process as `<TRestExpressionEvaluationProcess name="myVars" ... />`.

Then you observables will be written as `myVars_obsName`

jgalan:
I really find it useful that I am able to cut (or filter) any event using a condition at an analysisTree observable at **any** process, not only in your process. So, if you find the present approach has any limitations, the most appropriate would be to upgrade the present implementation to surpass those limitations.

I believe that is a parallel development. `TRestEvaluateExpressionProcess` should allow us to add new observables through the combination of observables, while *any* process should be able to add a *cut* or *filter*. Probably we may allow both words.

Vindaar:
> I really find it useful that I am able to cut (or filter) any event using a condition at an analysisTree observable at **any** process, not only in your process. So, if you find the present approach has any limitations, the most appropriate would be to upgrade the present implementation to surpass those limitations.
> 
> I believe that is a parallel development. `TRestEvaluateExpressionProcess` should allow us to add new observables through the combination of observables with mathematical operations, while _any_ process should be able to add a _cut_ or _filter_. Probably we may allow both keywords so everyone is happy.

Isn't the whole foundation REST that it runs a processing *pipeline*? In that vein doesn't it make sense to follow unix principles and have each tool do only one job, but do it well? As this process *only* does this thing, there's no harm in inserting it in between two other processes if some data filtering is required. Well, there's one downside, namely having to write a bunch of XML, but that's just the price to pay for having steering files.

But personally I don't care. I'm not going to remove sensible functionality from this process. If you feel like keeping a general `cut` functionality for all processes, feel free to use the expression evaluation library in the `TRestAnalysisTree`. As you say though, if one decides to go that way, this is independent of this PR.

On the other hand:
any idea why the observables do not show up, if I only use `SetObservableValue` instead of explicitly working with the analysis tree?

jgalan:
> On the other hand:
> any idea why the observables do not show up, if I only use `SetObservableValue` instead of explicitly working with the analysis tree?

My first guess is that at any REST process you need to define the observables explicitely using the `<observable` keyword. So, a process may have 100 observables implemented, but you are only interested into 1 of them. Then you just get that 1-observable at your analysisTree.

See for example how observables, and cut on the number of signals is implemented here.

https://github.com/rest-for-physics/framework/blob/master/pipeline/trex/01_raw.rml

The `observable` keyword is read by `TRestEventProcess::ReadObservables`, and added to the list of "enabled" observables.

So, for your process that makes probably a double definition. 

Any process is able to "enable" all implemented observables, so you may do the following in your process definition to make it work `<TRestExpressionEvaluationProcess ... observable="all" />` and that would do the job.

A better solution would be that you enable that in your process by default, i.e. that you don't need to write `observable="all"`.

I am not familiar with, but it seems that is carried out by `fDynamicObs` data member, as implemented at `LoadSectionMetadata`.

```
Int_t TRestEventProcess::LoadSectionMetadata() {
    TRestMetadata::LoadSectionMetadata();

    if (ToUpper(GetParameter("observable", "")) == "ALL") {
        fDynamicObs = true;
    }
...
```

So, probably finding a way to enable it at your process would do it. @nkx?

nkx111:
> I agree that duplication isn't a good thing. However, the supported expressions here are more powerful than those supported by <cut *:

You can make observables from the new `items` which have powerful expressions supported. Then add REST conventional `cuts` upon the observables. This avoids duplication.

```
///   <parameter name="observable" value="all"/>
///   <expressionset>
///     <!-- a simple, constant boolean expression -->
///     <item name="boolExpr" expr="5<10"/>
///     <!-- a simple, constant float expression -->
///     <item name="floatExpr" expr="5 * 10"/>
///     <!-- a float expression using an existing REST observable -->
///     <Item name="floatExprObs" expr="hitsAna_energy / 1000.0"/>
///     <!-- a bool expression using an existing REST observable -->
///     <item name="boolExprObs" expr="hitsAna_energy < 5000.0"/>
///     <!-- a bool expression of multiple statements combined by an `and` -->
///     <item name="boolAndExprMultiple" expr="hitsAna_energy < 5000.0 and tckAna_nTracks_X==1"/>
///     <!-- a bool expression of multiple statements combined by an `or` -->
///     <item name="boolOrExprMultiple" expr="hitsAna_energy < 5000.0 or tckAna_nTracks_X==1"/>
///     <!-- a complicated expression of nested parenthesis and multiple expressions -->
///     <item name="boolComplicated" expr="(hitsAna_energy / 1000.0) < 5.0 or (tckAna_nTracks_X==1 and tckAna_nTracks_Y==1)"/>
///     <!-- an expression used as a filter. I.e. if the predicate is true we keep the event, else drop it -->
///     <item filter="hitsAna_energy < 3000"/>
///     <!-- a mapping expression accessing a specific element from an observable that is map<int, Double> -->
///     <item name="veto4638" expr="peakTime[4638]"/>
///     <!-- a filtering expression accessing an observable that is map<int, Double> -->
///     <item name="peakTime" filter="peakTime[4638] < 300"/>
///   </expressionset>
///   <parameter name="cutsEnabled" value="on"/>
///   <cut name="hitsAna_energy" value=" < 3000"/>
///   <cut name="veto4638" value=" < 300"/>
```

> In addition the name cut is confusing, if things for which the predicate is true will be kept

It's just a positive/negative switch. Indeed it is more like a filter in current implementation. It is easy to modify.

> Finally, having a single process that is used to handle such a feature makes it clearer to use than having random "cuts" all over the place.

I think it good to have cuts all over the place. At least we should leave that possibility. The key point is that cutting bad events from the beginning saves process time.

jgalan:
> I think it good to have cuts all over the place. At least we should leave that possibility. The key point is that cutting bad events from the beginning saves process time.

I believe what @Vindaar means is that you would be able to place the `TRestExpressionEvaluationProcess` at any point on the chain.


Then, why do we need to encapsulate  the items between `<expressionset>`?

As soon as there is a proper validation test for this process, and a green badge at this PR, or another PR created from a new branch at `rest-for-physics/framework` I believe the process could be merged.

jgalan:
About the filter, I dont think it is interesting to have a **dedicated** process to remove events from a processing chain. People doing analysis will be able to apply advanced cuts/filters by accessing all the events from the final tree/RDataFrame. It is not a good idea to remove events from the chain. The best is to have all of them at the final tree, and then let decide people doing analysis how they will do the filtering. But **not** at the processing chain.

I was a bit not in favor already to introduce `cuts` (or filter) on each process basis. This might be a problem, at the end you dont get all events, and you never now if those events will have a particular interest for the people doing analysis, because you cannot foresee what the people doing analysis are searching for. So, in general, cutting data is not a good idea.

In brief, it is really simple, or it should, to filter data from the final event tree/RDataFrame without the need to cut/filter data on the event processing chain.

Having a process that we may connect anywhere to cut events in a, may be, crazy way is an idea I dont like. We should educate people using REST to do that only at the final tree, and not at the processing event chain.

Of course, if we need to introduce cuts at a given process because it is not worth to continue processing it, and it will add a big load to the processing chain, then in a very especial scenario, we can do that on each process using the `<cut` statement.

That way it will be clearer in which process we are applying cuts.

Still, we must also consider the scenario in which we do a quick event data processing, and then we use `TRestEventSelectionProcess` to launch the event processing only on those events we are interested in. While at the quick analysis we have all the events, **no** cuts. Implemented recently by @DavidDiezIb 

Plz, if not a big headache I invite you to remove the `filter` option. We just need to have the possibilty to add new variables to the final tree that will allow to do the filtering in a more straighforward way.

nkx111:
> The best is to have all of them at the final tree, and then let decide people doing analysis how they will do the filtering. But not at the processing chain.

I agree. And I believe this is what most collaborations are doing in their data chain. We don't need to struggle on event cut/filtering with processes.

On the other hand, evaluating expressions in process is indeed helpful. In PandaX-II data chain(not REST) we also calculate values for event quality cut, from the combinations of other values. Since the expression is hard-coded it is not convenient to adjust cut level. Here using rml could be much simpler.

Vindaar:
I'm just going to ignore the fact that you are contradicting yourself
multiple times here.

I *don't care* whether there is a global way to cut on events in REST. I'm
only highlighting the facts about the advantages that this expression
evaluation has over the existing solution, namely better
performance and more features.

Thinking that one could run an analysis chain without *any cuts
whatsoever* is in my opinion a bit foolish. For certain kinds of input
data that might be fine. In many cases you would however end up with
large amounts of data that would have to be analyzed for no purpose as
it would be thrown out at the end anyway.

It should be up to the user to decide what to keep and what to throw
out (because otherwise they will get creative and hack around the
missing functionality). Restricting functionality "I know better what
the user needs" leads to the iOSs and Androids of the world.

Further, removing functionality that can be approximated with worse
performance is insanity (computing a new variable that will be cut on
afterwards is wasted computation as we first run over all data once
and then run over it again to throw out elements). If you care about
the then obsolete feature, it should be in your interest to
allow that feature to be used in the global way instead of blocking
its introduction.

In addition all filters performed with this process are stored in meta
data, so the information is all there at the end. Yes, it would be
much better if meta data wasn't just a string, but a meaningful, structured data
type that could be worked with programmatically without parsing
strings. But also this: that's not my problem and it's unrelated to
this PR.

And having a single process that can do filtering at least makes it
obvious where filters are applied and one can check the meta data.

But seriously, I don't feel like outside involvement into REST is
appreciated, so I might as well stop here. Feel free to take it or
leave it. I fortunately don't have to care (I only know some people
would appreciate the functionality).

PR #73, opened 2021-06-21
Minor mesh updates

jgalan:
Large jgalan  [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/jgalan_minor_mesh_updates) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=rest-for-physics&utm_campaign=badge_info)

Simply added a method to be able to recover the energy at a given group nodes id (or cluster).

### DIFF
--- a/source/framework/core/inc/TRestMesh.h
+++ b/source/framework/core/inc/TRestMesh.h
@@ -60,16 +60,16 @@ class TRestMesh : public TObject {
     Int_t fNumberOfGroups = 0;
 
     /// A vector storing the group ID of the corresponding nodes activated
-    std::vector<Int_t> nodeGroupID;
+    std::vector<Int_t> fNodeGroupID;
     /// A vector storing the X-dimension cell id
-    std::vector<Int_t> nodeX;
+    std::vector<Int_t> fNodeX;
     /// A vector storing the Y-dimension cell id
-    std::vector<Int_t> nodeY;
+    std::vector<Int_t> fNodeY;
     /// A vector storing the Z-dimension cell id
-    std::vector<Int_t> nodeZ;
+    std::vector<Int_t> fNodeZ;
 
     /// A vector storing the total energy inside the cell id
-    std::vector<Double_t> energy;
+    std::vector<Double_t> fEnergy;
 
     /// A flag to indentify if we use cylindrical coordinates
     Bool_t fIsCylindrical = false;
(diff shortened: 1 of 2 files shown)


source/framework/core/src/TRestMesh.cxx
```diff
@@ -420,9 +420,22 @@ Int_t TRestMesh::GetGroupId(Double_t x, Double_t y, Double_t z) {
...
 
-    return nodeGroupID[index];
+    return fNodeGroupID[index];
+}
+
+///////////////////////////////////////////////
+/// \brief It returns the total energy of all nodes corresponding to the group id given by argument
+///
+Double_t TRestMesh::GetGroupIdEnergy(Int_t index) {
+    if (index > (int)fNodeGroupID.size()) return 0.0;
```
juanangp, line 432: This line is needed? Assuming GetNumberOfNodes() == 0, the lines below will just return 0...

source/framework/core/inc/TRestMesh.h
```diff
@@ -60,16 +60,16 @@ class TRestMesh : public TObject {
...
     /// A vector storing the Y-dimension cell id
-    std::vector<Int_t> nodeY;
+    std::vector<Int_t> fNodeY;
     /// A vector storing the Z-dimension cell id
-    std::vector<Int_t> nodeZ;
+    std::vector<Int_t> fNodeZ;
 
     /// A vector storing the total energy inside the cell id
-    std::vector<Double_t> energy;
+    std::vector<Double_t> fEnergy;
```
juanangp, line 72: Don't quite understand this design, you define different vectors with different parameters and then you assign the same index?
I would suggest to make a struct or class container:
```
class MeshParams {
Int_t ID;
Int_t fNodeX;
Int_t fNodeY;
Int_t fNodeZ;
Int_t fEnergy;
}
```
Then you just have to create an `std::vector<MeshParams>`

jgalan: Yes, that would be better, but right now I have no time for that ...

jgalan: Thats not this PR about. There are many other parts in the code where it would be good to have those kind of improvements.

jgalan: If you see that it is clearly an improvement to do it using a structure. Then, probably the best is that you create a new PR with those upgrades.

juanangp: If is spread everywhere in the code, I think you can create an issue about that.

jgalan: Sure, but before the issue for the structure we should make the issue for the `TRestMesh` validation routines, so if someone touches something at `TRestMesh` we make sure it doesn't break. xD

juanangp: If the vectors `std::vector<Double_t> fEnergy;` or similar are saved in a TTree, I guess it would break something or at least would not be backward compatible...

jgalan: Yes, it is a big change, not sure how good that would be accepted by ROOT schema evolution. But for the moment, there is no class storing `TRestMesh` on disk. It is only used as an auxiliary object, by `TRestAxionMagneticField` and `TRestFastHitsToTrackProcess`.

KonradAltenmueller (approved):

DavidDiezIb (approved):

PR #74, opened 2021-06-21
CMakeLists.txt update (compatibility with C++17)

lobis:
lobis  [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/cpp17) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=rest-for-physics&utm_campaign=badge_info)

The goal of this PR is to make REST compatible with C++17 (and not break anything in between).

I just updated the CMakeLists.txt file to set the CXX standard dynamically to match the one used to compiled ROOT.

I tried to do this using REGEX which are implemented in CMake but I failed, its not very well documented. Added TODO to the file.

```
# TODO: Simplify this with CMAKE STRING (REGEX ...) (https://cmake.org/cmake/help/latest/command/string.html)
```

Looks like this is enough to allow building REST with C++17 if ROOT was also built using C++17.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,8 +13,27 @@ elseif (CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
         message(FATAL_ERROR "Clang version must be at least 3.2!")
     endif ()
 endif ()
-set(CMAKE_CXX_STANDARD 14)
+
+find_package(ROOT REQUIRED)
+# TODO: Simplify this with CMAKE STRING (REGEX ...) (https://cmake.org/cmake/help/latest/command/string.html)
+execute_process(COMMAND root-config --has-cxx20 OUTPUT_VARIABLE HASCXX20)
+execute_process(COMMAND root-config --has-cxx17 OUTPUT_VARIABLE HASCXX17)
+execute_process(COMMAND root-config --has-cxx14 OUTPUT_VARIABLE HASCXX14)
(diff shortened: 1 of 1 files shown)


jgalan:
Nice! Still the validation pipeline will be running with the C++ version ROOT is compiled with.

Sometimes the compatibility is broken, so at least we should have 2 builds, one with C++14 and one with C++17. Just adding an additional `image:` instruction to new build task would be enough.

jgalan (approved):

juanangp (approved):
Nice feature.

Just for my understanding, it requires ROOT to be compatible with C++17/20 right?

jgalan:
> Nice feature.
> 
> Just for my understanding, it requires ROOT to be compatible with C++17/20 right?

As I understand it will compile REST c++17 if ROOT was compiled with c++17, REST c++20 if ROOT was compiled with c++20, etcetera. What I deduce from the file changes.

lobis:
> > Nice feature.
> > 
> > Just for my understanding, it requires ROOT to be compatible with C++17/20 right?
> 
> As I understand it will compile REST c++17 if ROOT was compiled with c++17, REST c++20 if ROOT was compiled with c++20, etcetera. What I deduce from the file changes.

Exactly, this is it.

It calls root-config after checking ROOT is installed. I don't think there is a way to get the c++ version root was compiled from a single command, so I did this ugly block of code. I couldn't get the regex match to work on cmake.

Issue #76, opened 2021-06-23
TRestReflector::ToString() does not correctly convert double


lobis:
https://github.com/rest-for-physics/framework/blob/master/source/framework/tools/src/TRestReflector.cxx

The current implementation:

```
string TRestReflector::ToString() {
    if (type == "string") return *(string*)(address);
    if (address == nullptr) return "null";
    RESTVirtualConverter* converter = RESTConverterMethodBase[type];
    if (converter != nullptr) {
        return converter->ToString(address);
    } else {
        return Form("Type: %s, Address: %p", type.c_str(), address);
    }
}
```

There is a problem when converting doubles such as timestamp. The output is truncated in such a way that the full number cannot be recovered. This is becaues the `cout` also doesn't display the full number, the `cout.precesion` needs to be bigger to correctly display this numbers.

PR #77, opened 2021-06-23
TRestReflector::ToString() for basic types now read cout precision

nkx111:
nkx111  [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/issue76) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=rest-for-physics&utm_campaign=badge_info)

This is a quick fix related to issue #76

Now TRestReflector::ToString() converts double/float to string with the same precision as we set for cout.

```
root [0] cout<<M_PI<<endl;
3.14159
root [1] any(M_PI).ToString()
(std::string) "3.14159"
root [2] cout.precision(16)
(long) 6
root [3] cout<<M_PI<<endl;
3.141592653589793
root [4] any(M_PI).ToString()
(std::string) "3.141592653589793"
```

### DIFF
--- a/source/framework/core/src/startup.cpp
+++ b/source/framework/core/src/startup.cpp
@@ -102,7 +102,10 @@ REST_Verbose_Level gVerbose = REST_Warning;
 // initialize converter methods
 template <class T>
 string ToStringSimple(T source) {
-    return ToString(source);
+    ostringstream ss1;
+    ss1.precision(cout.precision());
+    ss1 << source;
+    return ss1.str();
 }
 AddConverter(ToStringSimple, StringToInteger, int);
 AddConverter(ToStringSimple, StringToDouble, double);


lobis:
> nkx111  [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/issue76) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=rest-for-physics&utm_campaign=badge_info)
> 
> This is a quick fix related to issue #76
> 
> Now TRestReflector::ToString() converts double/float to string with the same precision as we set for cout.
> 
> ```
> root [0] cout<<M_PI<<endl;
> 3.14159
> root [1] any(M_PI).ToString()
> (std::string) "3.14159"
> root [2] cout.precision(16)
> (long) 6
> root [3] cout<<M_PI<<endl;
> 3.141592653589793
> root [4] any(M_PI).ToString()
> (std::string) "3.141592653589793"
> ```

Thanks for the quick reply! but insted of relying on the cout precison set by the user, shouldn't this method set (temporaly) a max precision so we don't lose information?

nkx111:
> but insted of relying on the cout precison set by the user, shouldn't this method set (temporaly) a max precision so we don't lose information?

In this way, there would be too many options. We may require precision for double/float, we may require scientific limits for int/long, we may require max number of output items for map, we may require formats for date/time. ToString() is working for any kind of classes and data types, so it is better to make it simple.

This method is inspired from https://docs.microsoft.com/en-us/dotnet/api/system.object.tostring?view=net-5.0

lobis (approved):

jgalan (approved):

PR #78, opened 2021-06-28
TRestDummyMetadata template class added

KonradAltenmueller:
Large KonradAltenmueller  [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/TRestDummyMetadata) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=rest-for-physics&utm_campaign=badge_info)

TRestDummyMetadata class added to serve as template for new Metadata classes

### DIFF
--- a/source/framework/core/inc/TRestDummyMetadata.h
+++ b/source/framework/core/inc/TRestDummyMetadata.h
@@ -32,6 +32,10 @@ class TRestDummyMetadata : public TRestMetadata {
     Double_t fDummy = 0;
     /// Dummy string
     TString fDummyString = "";
+    /// Dummy CONSTANT member
+    // It is better to use 'const' class members whenever possible (they won't be modified).
+    // These members do not need a Setter.
+    const Int_t fDummyConstNumber;  // const members can be initialized here, but... (see constructor)
 
    public:
     void Initialize();
@@ -42,7 +46,7 @@ class TRestDummyMetadata : public TRestMetadata {
     void PrintMetadata();
 
     // Constructor & Destructor
-    TRestDummyMetadata();
+    TRestDummyMetadata(Int_t dummyConstNumber);
     ~TRestDummyMetadata();
 
     // Setters & Getters
@@ -56,6 +60,8 @@ class TRestDummyMetadata : public TRestMetadata {
     // Avoid copying large objects!
     // All types, except built-in types (such as int, double...) should be passed via const reference
     inline void SetDummyString(const TString& dummyString) { fDummyString = dummyString; }
+
+    inline Int_t GetDummyConstNumber() const { return fDummyConstNumber; }
 
     ClassDef(TRestDummyMetadata, 1);
 };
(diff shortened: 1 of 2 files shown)


jgalan (approved):

nkx111 (commented):
It's better to add get/set methods to operate fDummy

lobis:
I also agree with @nkx111 that you should add getters and setters. Also I think all getters (and other methods whenever possible) should be `const` methods. This is a "tag" added to methods that do not modify any instance members, such as methods that only retrieve a value. This makes reading the code and navigating using an IDE easier.

jgalan:
I did a quick test

```
root [0] TRestDummyMetadata *m = new TRestDummyMetadata()
ROOT_prompt_0:1:29: error: allocating an object of abstract class type 'TRestDummyMetadata'
TRestDummyMetadata *m = new TRestDummyMetadata()
                            ^
/programas/REST/dev_javi/include/TRestMetadata.h:135:18: note: unimplemented pure virtual method 'InitFromConfigFile' in 'TRestDummyMetadata'
    virtual void InitFromConfigFile() = 0;
```

 it seems `InitFromConfigFile` needs to be implemented since it is directly inheriting from `TRestMetadata`. I wonder if `TRestEventProcess::ReadAllParams` could be promoted to `TRestMetadata` doing `InitFromConfigFile` not pure virtual anymore.

jgalan:
What is the need to introduce a const metadata member? A metadata class is usually initialised through a configuration file. Calling to the constructor that receives the config filename.

jgalan:
Still, it must have the problem that `InitFromConfigFile` is not implemented ...

nkx111:
You can implement `InitFromConfigFile` like:

```
void TRestDummyMetadata::InitFromConfigFile(){
  ReadAllParameters();
}
```

`ReadAllParameters()` is not defined in `TRestEventProcess`. It is defined from `TRestMetadata`. So any metadata classes can use it. We just don't use it by default execpt processes because metadata classes may have complex `InitFromConfigFile` logic.

jgalan:
Ok, then we could do that implementation directly at `TRestMetadata`?

```
virtual void TRestMetadata::InitFromConfigFile(){
  ReadAllParameters();
}
```

It will be overridden in case `InitFromConfigFile` is implemented.

In that implementation we would do:

```
void TRestDummyMetadata::InitFromConfigFile(){
  TRestMetadata::InitFromConfigFile();

  // do what-ever
}
```

nkx111:
> Ok, then we could do that implementation directly at TRestMetadata?

Yes, it is possible. And we shall still leave `TRestMetadata::InitFromConfigFile()` a pure virtual method, therefore keep `TRestMetadata` a abstract class

jgalan:
Yep, but if we implement `TRestMetadata::InitFromConfigFile` it is not anymore `pure virtual` but just `virtual`.

jgalan:
I guess this metadata class template is not necessary anymore, we may use the recently added `REST_MakeMetadata.C` macro to generate a metadata class skeleton.

PR #82, opened 2021-07-01
T rest map group

KonradAltenmueller:
- new metadata class TRestMapGroup
- allows to define a std::map<string,string> which can be added to an analysis file to facilitate offline analysis or plotting. In can contain for example calibration factors for different signal IDs, or one can use it to assign names to signal IDs.

### DIFF
--- a/source/framework/core/inc/TRestDummyMetadata.h
+++ b/source/framework/core/inc/TRestDummyMetadata.h
@@ -32,6 +32,10 @@ class TRestDummyMetadata : public TRestMetadata {
     Double_t fDummy = 0;
     /// Dummy string
     TString fDummyString = "";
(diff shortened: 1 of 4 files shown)


jgalan:
Hi @KonradAltenmueller I see just a minor problem in this PR.

The problem is that you are bringing to this PR commits belonging to another PR #78.

I believe what you could do is to create a new branch inheriting from master.

```
git checkout master
git fetch
git pull
```

Make sure you are in the latest 

```
git log
```

Then create a new branch for your commit

```
git checkout -b konrad_map_group
```

git cherry-pick 1a94175
git push
```

KonradAltenmueller:
Yes, I noticed. I am trying to fix it now...

PR #83, opened 2021-07-01
New metadata class TRestMapGroup

KonradAltenmueller:
KonradAltenmueller Large: 210 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/konrad_map_group) [](https://github.com/rest-for-physics/framework/commits/konrad_map_group)

- new metadata class TRestMapGroup
- allows to define a std::map<string,string> which can be added to an analysis file to facilitate offline analysis or plotting. It can contain for example calibration factors for different signal IDs, or one can use it to assign names to signal IDs.


source/framework/core/inc/TRestMapGroup.h
```diff
@@ -0,0 +1,58 @@
...
+ * If not, see http://www.gnu.org/licenses/.                             *
+ * For the list of contributors see $REST_PATH/CREDITS.                  *
+ *************************************************************************/
+
+#ifndef RestCore_TRestMapGroup
+#define RestCore_TRestMapGroup
+
+#include "TRestMetadata.h"
+
+/// The most simple metadata class to be used as a template
```
jgalan, line 28: I guess the brief metadata class description needs an update

source/framework/core/inc/TRestMapGroup.h
```diff
@@ -0,0 +1,58 @@
...
+class TRestMapGroup : public TRestMetadata {
+   protected:
+    /// Documentation of my dummy member
```
jgalan, line 31: Members should be documented. Use `///` on the line above for each metadata member.

source/framework/core/src/TRestMapGroup.cxx
```diff
@@ -0,0 +1,83 @@
...
+        groupDefinition = GetNextElement(groupDefinition);
+    }
+}
+
+///////////////////////////////////////////////
+/// Printing metadata member values on screen
+///
+void TRestMapGroup::PrintMetadata() {
+    TRestMetadata::PrintMetadata();
+
```
jgalan, line 75: What are the `fMapGroupKeys` and `fMapGroupValues` and why are not printed out?

source/framework/core/src/TRestMapGroup.cxx
```diff
@@ -0,0 +1,83 @@
...
+void TRestMapGroup::ConstructMap() {
+    for (unsigned int i = 0; i < fGroupKeys.size(); ++i) {
+        fMapGroup[fGroupKeys[i]] = fGroupValues[i];
+    }
+}
+///////////////////////////////////////////////
+/// \brief Set variables by default during initialization.
+///
+void TRestMapGroup::Initialize() { SetSectionName(this->ClassName()); }
+/// \brief Read Map Group from config file
```
jgalan, line 58: Format is inverted

jgalan (changes requested):

jgalan:
I thought the main purpose of this class would be to identify signal ids, i.e. integer numbers with a group name.

So, I would expect to have just a metadata member as:

```
std::map <string, std::vector<int>>;
```

where we store the group name, and the ids corresponding to that group.

The class then would be specific for signal ids, and perhaps the class could be renamed to something like `TRestMapGroupIds`.

KonradAltenmueller:
I thought it is more general to have a `map<string, string>`. For example for calibration factors we have `map<int,double>`, for the veto names `map<string,int>` or `map<int,string>`... And it is no big deal to convert the string in whatever type we want later.

Anyways, I still have trouble to launch this class from an rml file and wanted to make a post in the rest forum about it. When I get it to work I will of course document all members and so on and address you other comments.

jgalan:
Ok, but I think we are talking about two independent metadata classes. One to define sets of groups of ids, `TRestMapGroupIds` which I thought it was this PR --> `std::map< string, std::vector<int> >;` storing name + ids.

Then, a class to store the energy correction factors. This would be very similar to `TRestGainMap`. However, we do not implement a readout, so it would just serve to store correction factors.

```
std::map <int, double>;
```

source/framework/core/src/TRestMapGroup.cxx
KonradAltenmueller: They are two vectors with the map keys and values, read from the RML. Now they are printed out, too.

source/framework/core/src/TRestMapGroup.cxx
KonradAltenmueller: What do you mean?

KonradAltenmueller:
I don't think it is necessary to separate the class in `TRestMapGroupIds` and `TRestMapFactors` (but maybe it is better to change the name of this class).
If I use this class as it is to obtain a `map<string, string> vetoGroup` by defining in the RML 
```
<TRestMapGroup name="vetoGroup">
      <MapGroup key="veto_top" value="1,2,3"/> 
</TRestMapGroup>
```
the vector of Signal IDs (1,2,3) for each veto group I can obtain with `REST_StringHelper::StringToElements(vetoGroup["veto_top"])`. 
Currently, this seems to be good enough to me... Or did you have something else in mind?

source/framework/core/src/TRestMapGroup.cxx
jgalan: I mean, the `///////` should be before `/// \brief ....`

jgalan: The official format is as

```
////////////////////////
/// \brief ...
///
Int_t Method( xxx) {
```

jgalan: Leaving one blank line between methods

jgalan:
I understand you may use `StringToElements`, but then why dont just use string variables, and create methods for every format? There is an adantage to use c++ types.

jgalan:
> I don't think it is necessary to separate the class in `TRestMapGroupIds` and `TRestMapFactors` (but maybe it is better to change the name of this class). If I use this class as it is to obtain a `map<string, string> vetoGroup` by defining in the RML
> 
> ```
> <TRestMapGroup name="vetoGroup">
>       <MapGroup key="veto_top" value="1,2,3"/> 
> </TRestMapGroup>
> ```
> 
> the vector of Signal IDs (1,2,3) for each veto group I can obtain with `REST_StringHelper::StringToElements(vetoGroup["veto_top"])`. Currently, this seems to be good enough to me... Or did you have something else in mind?

Yes, but probably you need to implement a new method named `REST_StringHelper::StringToIntegerElements` so that you get back a `std::vector <int>`. You then fill the map with the group name and the vector. If the reply seems to be not accurate is because I didn't think about it with enough detail. Some times it happens to me.

PR #88, opened 2021-07-09
Event process updates

jgalan:
jgalan  [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/event_process_updates) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=rest-for-physics&utm_campaign=badge_info)

- Now we will be able to define a parameter process using any metadata class available to `TRestRun`. I.e. using `<parameter name="pressure" value="[TRestDetector::fPressure]" />`. In order for the process to be able to do such operation, it must do not implement `InitFromConfigFile`. Or if it does, it must call the one at `TRestEventProcess::InitFromConfigFile`

- The trexdm pipeline validation has been updated accordingly at `ValidateDetectorParams`.

- A shortcut introduced at `ReplaceMetadataMembers` so that if the string does not contain `[` it will just ignore the replacement.

- `ReplaceMetadataMember` is now private, it should only be called internally.

@rest-for-physics/core_dev


DavidDiezIb (approved):

nkx111 (approved):

nkx111 (commented):
It is very useful! Previously I just import those fields to REST_ARGS gloabl var and change `getparameter`

jgalan:
I see, I am not familiar with `REST_ARGS` and how it is used, but the present PR in principle should only affect to processes, through the update of `TRestEventProcess::InitFromConfigFile`.

So, how the code implementation at `TRestEventProcess` would affect the members being read at `TRestRun`?

Anyhow, the pipeline is running after commenting the lines at commit 8946b0f

nkx111:
REST_ARGS was designed initially to store arguments from command line. For example, `--i XXX.graw` will add inputFileName=XXX.graw in REST_ARGS. Last year, at some time, we were struggling on how to use TRestDetector and global variables. Global parameters were first stored in TRestDetector. Then after an online discussion we decided to save them somewhere else, and leave TRestDetector a metadata class. The new place to store global var was REST_ARGS. It was temporary. Now we defined a set of parameter-datamember conventions with reflection, and I think it is time to throw away global parameters in the code.

PR #94, opened 2021-07-20
Error notifications

jgalan:
Large jgalan  [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/error_notifications) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=rest-for-physics&utm_campaign=badge_info)

I am trying to create error notifications at the end of `restManager` execution. The idea behind is that we get a list of errors of processes or metadata structures that invoked `TRestMetadata::SetError`.

I thought implementing it at the end of `TRestProcessRunner::ConfigOutputFile()` would do the job. But it seems that only metadata classes directly added to `TRestRun` are available. But no processes.

For example, I added at the detector lib some notifications, 

https://github.com/rest-for-physics/detectorlib/pull/14

I added it for both the process and gas metadata, but only gas metadata error comes out. 

This is an example when I compile REST without Garfield, but I try to use `TRestDetectorGas`.

Screenshot 2021-07-20 at 23 18 40

### DIFF
--- a/source/bin/restManager.cxx
+++ b/source/bin/restManager.cxx
@@ -47,7 +47,7 @@ int fork_n_execute(string command) {
 }
 
 void PrintHelp() {
-    TRestStringOutput fout(COLOR_BOLDYELLOW, "", kHeaderedLeft);
+    TRestStringOutput fout(COLOR_BOLDYELLOW, "", kLeft);
     fout << " " << endl;
 
     fout.setheader("Usage1 : ./restManager ");
--- a/source/framework/core/inc/TRestMetadata.h
+++ b/source/framework/core/inc/TRestMetadata.h
@@ -174,19 +174,20 @@ class TRestMetadata : public TNamed {
     /// It can be used as a way to identify that something went wrong using SetError method.
     Bool_t fError = false;  //!
 
+    /// It counts the number of errors notified
+    Int_t fNErrors = 0;  //!
+
     /// It can be used as a way to identify that something went wrong using SetWarning method.
     Bool_t fWarning = false;  //!
+
+    /// It counts the number of warnings notified
+    Int_t fNWarnings = 0;  //!
 
     /// A string to store an optional error message through method SetError.
     TString fErrorMessage = "";  //!
 
     /// It can be used as a way to identify that something went wrong using SetWarning method.
     TString fWarningMessage = "";  //!
-
-    /// It allows to modify the error message. Only if SetError was called previously.
-    void SetErrorMessage(TString message) {
-        if (GetError()) fErrorMessage = message;
-    }
 
     std::map<string, string> GetParametersList();
     void ReadAllParameters();
@@ -202,12 +203,14 @@ class TRestMetadata : public TNamed {
     void SetError(TString message = "") {
         fError = true;
         fErrorMessage = message;
+        fNErrors++;
     }
 
     /// A metadata class may use this method to signal that something went wrong
     void SetWarning(TString message = "") {
         fWarning = true;
         fWarningMessage = message;
+        fNWarnings++;
     }
 
     /// Returns a string containing the error message
@@ -225,6 +228,10 @@ class TRestMetadata : public TNamed {
         else
             return "No warning!";
     }
+
+    Int_t GetNumberOfErrors() { return fNErrors; }
+
+    Int_t GetNumberOfWarnings() { return fNWarnings; }
 
     Int_t LoadConfigFromElement(TiXmlElement* eSectional, TiXmlElement* eGlobal,
                                 map<string, string> envs = {});
--- a/source/framework/core/inc/TRestMetadata.h
+++ b/source/framework/core/inc/TRestMetadata.h
@@ -177,12 +177,6 @@ class TRestMetadata : public TNamed {
     /// A string to store an optional error message through method SetError.
     TString fErrorMessage = "";  //!
 
-    /// A metadata class may use this method to signal that something went wrong
-    void SetError(TString message = "") {
-        fError = true;
-        fErrorMessage = message;
-    }
-
     /// It allows to modify the error message. Only if SetError was called previously.
     void SetErrorMessage(TString message) {
         if (GetError()) fErrorMessage = message;
@@ -194,6 +188,12 @@ class TRestMetadata : public TNamed {
    public:
     /// It returns true if an error was identified by a derived metadata class
     Bool_t GetError() { return fError; }
+
+    /// A metadata class may use this method to signal that something went wrong
+    void SetError(TString message = "") {
+        fError = true;
+        fErrorMessage = message;
+    }
 
     /// Returns a string containing the error message
     TString GetErrorMessage() {
--- a/source/framework/core/inc/TRestMetadata.h
+++ b/source/framework/core/inc/TRestMetadata.h
@@ -172,22 +172,22 @@ class TRestMetadata : public TNamed {
     map<string, string> fConstants;  //!
 
     /// It can be used as a way to identify that something went wrong using SetError method.
-    Bool_t fError = false;  //!
+    Bool_t fError = false;
 
     /// It counts the number of errors notified
-    Int_t fNErrors = 0;  //!
+    Int_t fNErrors = 0;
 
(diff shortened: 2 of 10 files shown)


jgalan:
Finally I added at TRestThread an output for process errors. Might not be the best, but any ideas welcome!

Screenshot 2021-07-21 at 09 41 06

jgalan:
Aesthetically I would prefer that the error output would come out as a whole, a kind of error summary report.

jgalan:
I added also warning messages in parallel. And counters to now the times the `SetError` and `SetWarning` methods were invoked.

jgalan:
I have now changed the `TRestMetadata` members so that they will be written to disk.

Then, I implemented methods at `TRestRun` so that we will be able to print errors/warnings that happened during the processing.

Screenshot 2021-07-22 at 07 37 48

jgalan:
I was thinking it would be also possible to upgrade those metadata members so that we are allowed to define different error types.

`TString fErrorMessage;`  becoming `std::map<string,string> fErrorMessage;` //type and error message

And `std::map<string,int> fNErrors;` //type and error count

Since `TRestMetadata` structure already change, perhaps it would be good to decide this now.

@rest-for-physics/core_dev

jgalan:
Since we have now `run0->PrintErrors` and `run0->PrintWarnings` available, we could avoid printing the errors output on screen, and just notify that there were a number of errors, and those can be checked using `run0->PrintErrors()`.

nkx111:
Collecting errors/warnings at the end of process is useful. But I think it is better to set errors together with `ferr` message output. Then the modification of code could be minimal. I think modifying the struct `endl_t` would also do the job. Then, we don't need to call `SetError()` in the processes.

nkx111:
It may be good to also place error messages in a global variable, instead of individual metadata classes.

jgalan:
I thought it would be good to associate it with the metadata class itself. Dont know how it works the `messageBuffer` you mention.

I wrote few guidelines of this implementation here

https://rest-for-physics.github.io/rest-advanced/process-development.html

If it would be something more advanced, then I would say it would be good to encapsulate that in a new class `TRestLogger` that serves to register these messages. Perhaps using an existing logger. But that it is not straightforward and may take more work. Perhaps @lobis knows about good logging systems.

lobis:
> I thought it would be good to associate it with the metadata class itself. Dont know how it works the `messageBuffer` you mention.
> 
> I wrote few guidelines of this implementation here
> 
> https://rest-for-physics.github.io/rest-advanced/process-development.html
> 
> If it would be something more advanced, then I would say it would be good to encapsulate that in a new class `TRestLogger` that serves to register these messages. Perhaps using an existing logger. But that it is not straightforward and may take more work. Perhaps @lobis knows about good logging systems.

I have played around with `spdlog`, a good cpp logging library (https://github.com/gabime/spdlog). You can see implementation details in the https://github.com/rest-for-physics/restGeant4 project. It is capably of doing threadsafe logging, traceback and logging to a file etc, things we very much need.

nkx111:
> Dont know how it works the messageBuffer you mention.

All the ouputs through `fout`, `ferr`, `warning`, `essential`, `info`, `metadata`, `success` provided by TRestStringOutput will be saved to the `messageBuffer` of the coresponding metadata object. This is already implemented. So actually saving error and warning messages to `fErrorMessage` and `fWarningMessage` after printing them is a repeated work. 

Although in the new code of `TRestDetectorHitsToSignalProcess` it seems the error messages are not printed during run time.

jgalan:
Logging too much might slow down the processing. But any contributions directly to REST, in that sense are welcome.

Anyhow, the present implemention at this PR is simple enough to track user errors that happened during processing.

Please, could you approve this PR and the corresponding one at detectorlib?

lobis (approved):

lobis:
> Logging too much might slow down the processing. But any contributions directly to REST, in that sense are welcome.
> 
> Anyhow, the present implemention at this PR is simple enough to track user errors that happened during processing.
> 
> Please, could you approve this PR and the corresponding one at detectorlib?

That is preciesly why we need a dedicated logging library. Currently there is a trade-off between information and speed. Using a logging library with backtrace support would solve this issue, output logs would only be printed when they are relevant, but they are not lost when they are not printed. For example, we may want to print all information about steps in a simulation using the debug mode, but only for "interesting" events, so we keep the logs in a buffer and print them only if they are needed.

jgalan:
> All the ouput messages through `fout`, `ferr`, `warning`, `essential`, `info`, `metadata`, `success` provided by `TRestStringOutput` will be saved to `messageBuffer` of the coresponding metadata object.

Ok I didnt know that. It is more like a logger, than a punctual error notification. I believe the implementation at this PR might serve to notify of errors and warnings that are kind of crucial to the user to know that a kind of problem might be going on with the results he will get. In that sense, to produce a list of warnings and errors as `TRestRun::PrintWarnings` or `TRestRun::PrintErrors` will be useful for those users not willing to enter too much into all the output. Thats the main idea of this PR.

jgalan:
> Although in the new code of `TRestDetectorHitsToSignalProcess` it seems the error messages are not printed during run time. Then it is not repeated. But is it good to show errors after the process, as a whole? Do we have so many errors?

Yes, the idea is to attract the attention of users on the fact that an error happened. Do we have so many errors? Well, it might happen that when you process 100 events we find a percentage of those events something goes probably wrong. So, the user will know on how many of those events this problem occurs which is crucial to decide to take action and fix something, or are marginal events.

Of course, for the moment I only implemented at `TRestDetectorSignalToHitsProcess` but all processes might define critical error messages or warnings using `SetError` and `SetWarning`.

nkx111:
> It is more like a logger, than a punctual error notification.

We can also modify this logger to store error/warning messages for `ferr` and `warning` object. This is more integrated than individual `SetError()` method.

I suggest to not use `SetError/SetWarning`, or make them private which are called only by `TRestStringOutput`.

jgalan:
somehow I like the idea that we decide when to give certain relevance to an error message inside a process, meaning that the user should be prevented of certain anomaly. Thats why I think implementing explicitely `SetError/SetWarning` in our processes will be useful as it is now at this PR. Of course, having a parallel logging with everything is more appropriate for developers.

jgalan:
> I suggest to not use `SetError/SetWarning` explicitly. Maybe we can make them private and only call in `TRestStringOutput::ferr` and `TRestStringOutput::warning`.

Yes, if we can achieve calling `SetError` and `SetWarning` there it would be also a good solution. We could then decide to show all the warnings/errors during the execution, or just a notification message saying the user that there was errros and s/he should use `run0->PrintErrors()` or `mdName->PrintMessageBuffer()`.

nkx111:
OK, I will try to integrate `SetError/SetWarning` inside `ferr/warning`.

nkx111:
It should be working now. We can use `ferr` and `warning` to pring while saving error messages as metadata.

nkx111:
It should be working now. We can use `ferr` and `warning` to print while saving those messages as metadata

jgalan:
Hi @nkx, the problem is that after the updates I do not get any warnings. The good point of error/warning notifications is to make aware the user of any problem even if they did not ask to know about it.

Screenshot 2021-07-27 at 13 32 10

See how it was working before your commits, in the last screenshots at this PR

https://github.com/rest-for-physics/tracklib/pull/8

nkx111:
The warning summary message are shown on my side. Without process's event level warnings. I use `warning<<endl;` in `TRestDetectorSignalToHitsProcess`. I use the given rml file from the forum. 
http://rest-forum.unizar.es/t/track-analysis-process-produces-a-seg-fault/500/6

image

But if we use `SetWarning()` in the process, there is indeed some problems. I have fixed that.

jgalan:
> But if we use `SetWarning()` in the process, there is indeed some problems. I have fixed that.

What fix? After updating to the latest update at this PR I see no effect.

Dont you need to push some changes here? rest-for-physics/detectorlib#14

jgalan:
Ok, perfect. It works on my side. Shall we merge as it is now?

nkx111 (approved):

PR #99, opened 2021-07-30
Analysis plot fix

jgalan:
Large jgalan  [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/analysis_plot_fix) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=rest-for-physics&utm_campaign=badge_info)

- A fix when using a full-path filename at `save=""`.
- A fix when getting run info.
- A new validation for analysisPlot when using `classify fRunNumber=`.

I identified a problem on `TRestAnalysisPlot::GetRunInfo`. In some circumstances the run number for the corresponding file was not being properly retrieved. It seems that `GetRunInfo` optimises the opening/closing of files, in case the `fRun` is already open, however, there were some problems. The fix I introduced is bypassing that optimisation and opening the file each time this method is called.

Of course, this is not the best, but as it is right now it works, and I added a new validation test for analysisPlot to check this problem.

If these lines are removed,

```
571    fRun->OpenInputFile(fileName);
572    return fRun;
```

the validation `ValidateClassify.C` will fail.

Any quick fix to solve this? @nkx?

### DIFF
--- a/source/framework/core/src/TRestAnalysisPlot.cxx
+++ b/source/framework/core/src/TRestAnalysisPlot.cxx
@@ -550,7 +550,7 @@ Int_t TRestAnalysisPlot::GetPlotIndex(TString plotName) {
 }
 
 TRestAnalysisTree* TRestAnalysisPlot::GetTree(TString fileName) {
-    if (fileName == fRun->GetInputFileName(0)) {
+    if (fRun->GetInputFile() != NULL && fRun->GetInputFile()->GetName() == fileName){
         // this means the file is already opened by TRestRun
         return fRun->GetAnalysisTree();
     }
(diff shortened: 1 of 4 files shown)


source/framework/core/src/TRestAnalysisPlot.cxx
```diff
@@ -562,6 +568,9 @@ TRestAnalysisTree* TRestAnalysisPlot::GetTree(TString fileName) {
 }
 
 TRestRun* TRestAnalysisPlot::GetRunInfo(TString fileName) {
+    fRun->OpenInputFile(fileName);
```
juanangp, line 571: This doesn't looks fine to me.

We can consider to move to singleton:
https://stackoverflow.com/questions/1008019/c-singleton-design-pattern

But I am afraid it would require a lot of changes.

jgalan: Ok, I am not familiar with. 

what I am worried now is that pandaxiii pipeline did not succeed.

nkx111 (approved):

nkx111:
Both pipelines are fixed and working now.

jgalan:
@juanangp will you approve?

juanangp (approved):

PR #101, opened 2021-08-19
TRestThread: fix a bug in non-test-run mode when last process is general process

nkx111:
If we disable test run, and if the last process is general process, then it will come into problem trying to assembly TRestEvent abstract class.


jgalan:
This PR is not passing the pipeline

nkx111:
It may because the base is too old. I merged from master and now the pipeline is passed.

jgalan (approved):

PR #107, opened 2021-10-08
Removed overwriting of CMAKE_INSTALL_PREFIX when it has been defined …

lobis:
lobis  [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/lobis-cmakelists)

…by user

CMake should not change the user defined variable CMAKE_INSTALL_PREFIX if it has been defined by the user.

For example, previous to this change, you could not install the package `restG4` via setting CMAKE_INSTALL_PREFIX as it would be installed to REST_PATH instead, making CMAKE_INSTALL_PREFIX not work as intended.

I don't think this change breaks anything but please let me know if you think it does.

@rest-for-physics/core_dev


jgalan:
As it is described at the installation

https://rest-for-physics.github.io/installation.html

```
cd ~/rest-framework
mkdir build
cd build
cmake .. -DINSTALL_PREFIX=../install/master/ 
make -j4 install
```

I agree that it could be changed to `CMAKE_INSTALL_PREFIX`. However, then, please, change the installation instructions accordingly.

nkx111:
I agree we should use `-DCMAKE_INSTALL_PREFIX` from now on. The changes shall be OK since it only containes renaming.

Issue #109, opened 2021-11-08
Introduce a testing framework


lobis:
Do the necessary steps to introduce a testing framework to better test our code (it currently only runs integration tests).

The proposed testing framework of choice is [Google Test](https://github.com/google/googletest).

PR #110, opened 2021-11-08
Google Test dependency in framework

lobis:
Related to https://github.com/rest-for-physics/framework/issues/109.

Added Google Test as a dependency (https://github.com/google/googletest). It uses the CMake Fetch functionality to automatically download the necessary files so there is no need to bundle the source code in the framework as external source.

This is the way to support the testing framework according to the documentation, probably it is possible to do it the old way but I think this is easier.

The drawback is that we need to update CMake version to at least 3.14 (latest is 3.22). Currently REST CMake minimum version is 3.5, so this may be problematic for some users with very old cmake version installed.


jgalan:
Notice that this branch is not synced with master

lobis:
> Notice that this branch is not synced with master

Fixed, thanks.

lobis:
@nkx111 Currently the docker image used in the pipeline uses CMake 3.5 which is too low for these changes. I think it would be a good idea to upgrade the CMake version to the latest stable version. It is possible that these changes can be implemented in a old CMake version but I don't know how, feel free to do a commit if you know how.

PR #111, opened 2021-11-09
Changed Docker image for GitLab CI pipeline, cmake tweaks

lobis:
Large lobis  [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/gitlab-ci-dockerimage)

New image uses more recent build tools such as cmake 3.16 which is needed by #109. It should work as the previous one.


juanangp:
Pipeline is faling, I guess we should fix the cmake to find Gardfield.

juanangp:
I did a work around to avoid the Garfield cmake error, it worked but the pipeline still failing, looks like a compilation error.

lobis:
I think I have fixed this issue, the problem with Garfield is solved but it breaks backwards compatibility with older versions of Garfield due to classes having been renamed. I think its fair to drop this old version (~5 years old) and stick with the "new" one. I guess we could maintain backwards compatibility with clever use of macros, if someone wants to make a commit.

The way Garfield is linked now is not the prettiest one, in fact, its pretty ugly. I tried to do it properly but couldn't due to the complex structure of our build. I think its time to refactor the cmake code. I will do this down the line.

Also in this PR we changed the CI Docker image to a newer one. This is the image I use for development and so far it has worked pretty well. I have automated the production of new tags so I can get a different C++ standard / ROOT or Geant4 version without much effort.

lobis:
Now the pipeline has failed due to a dictionary problem. I will try to fix it.

jgalan:
I have seen that problem before, but not 100% sure of what it is the origin. Perhaps it is in fact that Garfield `pcm` files are not accessible to the installation? They should be placed at `$GARFIELD_INSTALL/lib`, where `$GARFIELD_INSTALL/lib` is included at `LD_LIBRARY_PATH`.

Hope that helps

lobis:
It looks like commit [01f7d13](https://github.com/rest-for-physics/framework/pull/111/commits/01f7d139d31a466f35c7c18ff1c973c7eea09071) fixed the pipeline issue. In this commit I added a flag (-q) to quit instead of launching the root REPL. In principle this should not be needed, as the -q flag should be understood without explicitly defining it, and this seems to be the case up until now.

For some reason on the GitLab job this fails, but not when testing it manually on the Dockerfile. I have no idea why this is the case but this commit fixes it. I don't think this has any side-effects.

source/bin/restRoot.cxx
```diff
@@ -16,7 +16,6 @@ int main(int argc, char* argv[]) {
     setenv("REST_VERSION", REST_RELEASE, 1);
 
     Int_t loadMacros = 0;
-    bool quit = false;
```
jgalan: why coming back to previous implementation?

PR #113, opened 2021-12-05
Change directory

jgalan:
jgalan  [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/change_directory)

Adds two new methods at TRestTools, `ChangeDirectory` and `ReturnToPreviousDirectory`.

Those were implemented in order to be used at `restG4`. rest-for-physics/restG4#27

@rest-for-physics/core_dev

### DIFF
--- a/source/framework/core/src/TRestHits.cxx
+++ b/source/framework/core/src/TRestHits.cxx
@@ -551,9 +551,9 @@ void TRestHits::WriteHitsToTextFile(TString filename) {
     FILE* fff = fopen(filename.Data(), "w");
     for (int i = 0; i < GetNumberOfHits(); i++) {
         if ((fType.size() == 0 ? !IsNaN(fX[i]) : fType[i] % X == 0))
-            fprintf(fff, "%d\t%e\t%e\t%e\t%e\n", i, fX[i], "NaN", fZ[i], fEnergy[i]);
+            fprintf(fff, "%d\t%e\t%s\t%e\t%e\n", i, fX[i], "NaN", fZ[i], fEnergy[i]);
         if ((fType.size() == 0 ? !IsNaN(fY[i]) : fType[i] % Y == 0))		
-            fprintf(fff, "%d\t%e\t%e\t%e\t%e\n", i, "NaN", fY[i], fZ[i], fEnergy[i]);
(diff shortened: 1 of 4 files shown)


jgalan:
@nkx I had a compilation issue that I solved at commit bab19d1cc844c33ce8880797f3af3a66bfd59c8a 

Perhaps it is connected to, or required by, Microsoft Windows compilation?

nkx111:
There was `chdir()` used in TRestGDMLParser and DetectorConstruction in order to load gdml file correctly. (gdml must be loaded from the current directory). I think we better change directory in these class level, not in restG4 level.

These methods can be kept and called in TRestGDMLParser and DetectorConstruction to replace old `chdir()` method.

jgalan:
Right, in fact, the main problem is about geometry, but in future, if we define other relative paths/filenames, it is not better that the main program moves to the RML directory?

However, the argument of restG4 may receive a relative path `restG4 path/relative/to/rml/location/file.rml` and then the geometry might be relative to RML location. Then, we need to transport somehow this relative path `path/relative/to/rml/location/` to TRestGdmlParser` and `DetectorConstruction`. Isn't?

nkx111:
> Then, I believe the approach you mention would be easily implemented at TRestGdmlParser and DetectorConstruction by adding the relative path?

I mean we should call from certain class, not in restG4, the new methods `TRestTools::ChangeDirectory()`, `TRestTools::ReturnToPreviousDirectory()` introduced in this PR. 

That's a personal flavor, I would like to maintain old directory as long as possible, to avoid being confused where I am, or causing strange problems. For example, if we change directory in restG4 (https://github.com/rest-for-physics/restG4/pull/27), then when writing/modifing  `TRestGeant4Metadata`(e.g., save debug logs to an output file), I shall consider what directory I am saving file to. If is called in restG4 then it saves to gdml path. If is called in restRoot then it saved to current dir. This is confusing. Or if we happen to call `ReturnToPreviousDirectory()` twice due to some bugs, it is very hard to notice.

jgalan:
Anyhow, it seems the problem is at rest-for-physics/restG4#27, I guess adding these methods at `TRestTools` will not harm. Do we merge this PR and keep discussing at restG4 side?

nkx111 (approved):

PR #116, opened 2021-12-09
TRestMetadata: undefine = operator method overload

nkx111:
nkx111  [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/nkx111-patch-2)

It will cause double free problems if we use `=` or copy constructor for metadata classes. The reason is that the data member `fElement` as a pointer will also be copied. And when we delete the two metadata classes, the same pointer will be deleted twice. For example:

```cpp
TRestDetectorGas *g=new TRestDetectorGas("server","Xenon-TMA 0.1Pct 10-10E3Vcm") // initialize a metadata
{TRestDetectorGas gg = *g;} // make a new copy and delete.
delete g;  // this delete operation will cause seg.fault
```

In this PR I undefined the `=` operator. Now the second line doesn't work, and avoids this problem.

Now to copy metadata or to use the object(instead of pointer), one can use `Clone()`, or reference:

```cpp
TRestDetectorGas* g=new TRestDetectorGas("server","Xenon-TMA 0.1Pct 10-10E3Vcm") 
TRestDetectorGas* gg = (TRestDetectorGas*)g->Clone();
delete g; 
delete gg;
```

```cpp
TRestDetectorGas *g=new TRestDetectorGas("server","Xenon-TMA 0.1Pct 10-10E3Vcm")
{TRestDetectorGas& gg = *g;}
delete g;
```

### DIFF
--- a/pipeline/metadata/readout/PrintReadout.C
+++ b/pipeline/metadata/readout/PrintReadout.C
@@ -33,7 +33,7 @@ void PrintReadout(TString fName) {
     while ((key = (TKey*)nextkey())) {
         if (((string)(key->GetClassName())).find("TRestDetectorReadout") != -1) {
             TObject* obj = f->Get(key->GetName());
-            TRestDetectorReadout readout = *(TRestDetectorReadout*)obj;
+            TRestDetectorReadout& readout = *(TRestDetectorReadout*)obj;
             readout[0].Print(1);
         }
         n++;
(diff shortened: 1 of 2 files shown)


lobis (approved):

jgalan:
Hi @lobis, it is great you took your time to review and approve this PR. However, we should wait till the pipeline is green. So I just requested a second review.

jgalan:
I have created a new PR at detector library to try to fix the pipeline rest-for-physics/detectorlib#29

jgalan (approved):

PR #122, opened 2022-01-26
Alphas example

jgalan:
Large jgalan  [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/alphas_example)

Adding a new example showing a basic alpha track processing scheme.

See the [README.md](https://github.com/rest-for-physics/framework/tree/alphas_example/examples/01.alphaTrack) from the example for more details.

Have a look also to the corresponding [restG4 example README.md](https://github.com/rest-for-physics/restG4/tree/alphas_example/examples/08.Alphas).

@rest-for-physics/core_dev


nkx111 (approved):

juanangp (approved):

jgalan:
Pipeline is failing, so in principle the PR should not be accepted.

jgalan:
I am trying to look at the generateReadout.rml validation. Locally it fails because it tries to look for a decoding file under `/usr/local/REST/master/data/decoding/MM_V2_On_AGET0-1.dec`.

I have seen at the `REST_PATH/data/decoding/` it exists `MM_On_AGET0-1.dec` but it does not exist the `V2` version. 

I wanted to regenerate the `validation.txt` file in order to update the validation, since now an additional member is printed out when invoking `PrintMetadata()`.

nkx111:
I think we should not do validation by comparing to txt files. There is too many version/system dependent things when printing. I think we can just call some methods to see if the returning value is correct. For example, get channel Id from daq, get channel 0's position, get total number of channels, etc.

jgalan:
Yes, I agree with that, I thought it was you who introduced txt comparison. Also there are places in the pipeline where png binary images are compared directly.

So, for the moment I just remove the txt comparison.

But what happens with the /usr/local/ file?

Issue #125, opened 2022-02-03
Refactor CMake build


lobis:
We need to clean the CMake code as it is too old and complicated.

lobis:
https://github.com/rest-for-physics/framework/pull/145 partially solves this

PR #129, opened 2022-02-04
Support for testing framework

lobis:
Large lobis  [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/cmake-ctest)

Related to https://github.com/rest-for-physics/framework/issues/125.

The goal of this PR is to clean the CMake build system wherever possible (it is really complex, it could probably be simplified) and also modify it to support unit testing via ctest. Also I would like to install external dependencies via cmake instead of just copying them into the framework (such as tinyxml) but I won't change tinyxml since its been this way forever, if anything we should probably remove this dependency altogether and replace it by the ROOT xml library which can do what we want.

Testing is enabled via the `-DTEST=ON` flag, disabled by default.

Tests can be run via ´ctest´ from the build directory.

I will write the boilerplate for testing in all modules. (https://github.com/rest-for-physics/framework/issues/163).

Some information about running tests can be found at https://github.com/rest-for-physics/rest-for-physics.github.io/commit/43a1e66b464fbad98f3772e19654e9e440ab33fe.


jgalan:
As we get more and more activity on the repositories we should be more and more explicit on describing the intention of the updates at a specific PR. Please, give details on the problem and the enhancements that will bring this PR, including the reasons that triggered it. Those habits will be good to incentive discussions on the different contributions.

lobis:
Currently there is a problem I don't know how to fix, perhaps @nkx111 or @jgalan can give me a hand. It has to do with the `startup.cpp` script.

Currently I cannot compile the tests due to an error at the end of compilation:

```
sh: 1: rest-config: not found
CMake Error at /usr/share/cmake-3.16/Modules/GoogleTestAddTests.cmake:40 (message):
  Error running test executable.

    Path: '/tmp/tmp.1ggMShb7jT/cmake-build-dockerdev/source/framework/test/testFramework'
    Result: Child aborted
    Output:
      REST ERROR!! Lacking system env "REST_PATH"! Cannot start!
      You need to source "thisREST.sh" first
```

To get the error just compile REST with the new TEST flag on: `cmake -DTEST=ON ..`. 

I am not sure how/why a script is being run at compilation but for the all the tests targets I would like to avoid it, but don't know how to do this without probably breaking something since I don't understand this part of the code.

I think the implementation of the tests is good and I can't think of a modification that would avoid this so we probably need to modify `startup.cpp` or some cmake section that is "linking" this?

Thanks!

nkx111:
> Currently there is a problem I don't know how to fix, perhaps @nkx111 or @jgalan can give me a hand. It has to do with the `startup.cpp` script.
> 
> Currently I cannot compile the tests due to an error at the end of compilation:
> 
> ```
> sh: 1: rest-config: not found
> CMake Error at /usr/share/cmake-3.16/Modules/GoogleTestAddTests.cmake:40 (message):
>   Error running test executable.
> 
>     Path: '/tmp/tmp.1ggMShb7jT/cmake-build-dockerdev/source/framework/test/testFramework'
>     Result: Child aborted
>     Output:
>       REST ERROR!! Lacking system env "REST_PATH"! Cannot start!
>       You need to source "thisREST.sh" first
> ```
> 
> To get the error just compile REST with the new TEST flag on: `cmake -DTEST=ON ..`.
> 
> I am not sure how/why a script is being run at compilation but for the all the tests targets I would like to avoid it, but don't know how to do this without probably breaking something since I don't understand this part of the code. This is blocking testing right now :(
> 
> I think the implementation of the tests is good and I can't think of a modification that would avoid this so we probably need to modify `startup.cpp` or some cmake section that is "linking" this?
> 
> Thanks!



It seems that the new [ExampleTest.cxx](https://github.com/rest-for-physics/framework/pull/129/files#diff-9a8983ea7c7124a3610643bbd6b2c0ae4ac9ceca68cb6a2377a1ec582d17f7bd) is being executed and raised this error. The part of error code from startup.cpp will run as long as the framework library is linked.

lobis:
> > Currently there is a problem I don't know how to fix, perhaps @nkx111 or @jgalan can give me a hand. It has to do with the `startup.cpp` script.
> > Currently I cannot compile the tests due to an error at the end of compilation:
> > ```
> > sh: 1: rest-config: not found
> > CMake Error at /usr/share/cmake-3.16/Modules/GoogleTestAddTests.cmake:40 (message):
> >   Error running test executable.
> > 
> >     Path: '/tmp/tmp.1ggMShb7jT/cmake-build-dockerdev/source/framework/test/testFramework'
> >     Result: Child aborted
> >     Output:
> >       REST ERROR!! Lacking system env "REST_PATH"! Cannot start!
> >       You need to source "thisREST.sh" first
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > To get the error just compile REST with the new TEST flag on: `cmake -DTEST=ON ..`.
> > I am not sure how/why a script is being run at compilation but for the all the tests targets I would like to avoid it, but don't know how to do this without probably breaking something since I don't understand this part of the code. This is blocking testing right now :(
> > I think the implementation of the tests is good and I can't think of a modification that would avoid this so we probably need to modify `startup.cpp` or some cmake section that is "linking" this?
> > Thanks!
> 
> It seems that the new [ExampleTest.cxx](https://github.com/rest-for-physics/framework/pull/129/files#diff-9a8983ea7c7124a3610643bbd6b2c0ae4ac9ceca68cb6a2377a1ec582d17f7bd) is being executed and raised this error. The part of error code is from startup.cpp and will run as long as the framework library is linked.
> 
> I suggest not to avoid linking it. We can try adding those envs to gtest system.

but how is it being executed on the build? How could we add this env to the gtest in a simple way? are you sure running some code when linking to REST library is a good practise? seems it can cause a lot of trouble.

lobis:
something very weird happened with this branch I cannot fix the bug

jgalan:
> Also I would like to install external dependencies via cmake instead of just copying them into the framework (such as tinyxml)

In principle, there is no problem to do that with the present cmake system, if `tinyxml` would be installed as a system library it would just pick up it. I believe @nkx introduced some modifications to the library, so thats why the library is copied into the repository. Not because it is not possible to link external libraries.

CMakeLists.txt
```diff
@@ -1,4 +1,5 @@
-PROJECT(REST)
+PROJECT(REST LANGUAGES CXX HOMEPAGE_URL https://github.com/rest-for-physics/framework)
```
jgalan, line 1: Perhaps the home page could be https://rest-for-physics.github.io

source/framework/core/src/startup.cpp
```diff
@@ -25,6 +25,10 @@ map<string, RESTVirtualConverter*> RESTConverterMethodBase = {};
 struct __REST_CONST_INIT {
    public:
     __REST_CONST_INIT() {
+// TODO: fix to avoid this dirty fix
```
lobis, line 28: @nkx111 this is what I currently do to support testing, but its very ugly, perhaps you have some better idea?

With this "fix" if you build and install REST with testing enabled, perhaps it won't work as intended due to this line. It has no effect if testing is not enabled (default).

jgalan: How do you enable testing? Perhaps you should provide more details on the PR description so that we are able to reproduce what it is included at this PR

lobis:
This is ready for review and merge @jgalan @nkx111 @juanangp

juanangp:
Perhaps you should mention somewhere that the support for old Garfield versions have been removed, since you removed `FindGarfieldOld`, therefore is related with https://github.com/rest-for-physics/detectorlib/pull/41

lobis:
> Perhaps you should mention somewhere that the support for old Garfield versions have been removed, since you removed `FindGarfieldOld`, therefore is related with [rest-for-physics/detectorlib#41](https://github.com/rest-for-physics/detectorlib/pull/41)

I decided to keep this file and remove it in a more appropriate PR.

source/framework/test/files/metadata.rml
```diff
@@ -0,0 +1 @@
+<metadata  p1="75"  p2="12.32"  p3="Aloha" />
```
jgalan, line 1: It does not keep the convention that the xml header is the name of the c++ metadata class, which is `metadataTestClass`.

source/framework/test/src/FrameworkCore.cxx
```diff
@@ -37,3 +39,26 @@ TEST(FrameworkCore, TRestRun) {
     EXPECT_TRUE(run.GetRunDescription() == "This is a test for TRestRun");
     EXPECT_TRUE(run.GetVerboseLevelString() == "debug");
 }
+
+TEST(FrameworkCore, TRestMetadata) {
+    // Check required files exist
+    EXPECT_TRUE(fs::exists(BASIC_TRESTMETADATA_RML));
+    
+    // Create new TRestMetadata class
+    class metadataTestClass: public TRestMetadata{
```
jgalan, line 48: Any REST class should be starting by `TRest`

lobis: > Any REST class should be starting by `TRest`

but this is not a class, its just a cpp file

jgalan: well, it is a class, as it is stated in front, I agree we could make an exception inside tests, but I though it was not possible that the XML header and the class name would be different, it is this possible? @nkx?

lobis: what are you saying is a class? I'm not following you @jgalan

DavidDiezIb: I don't mind the naming convention, we can change it if you prefer. Just I thought it was better not to use REST naming convenctions to avoid confussion with any real REST classes. 
This is a fake one created to be able to test `GetParameter` function from TRestMetadata. It is a pure abstract class so we can't call it, we can only inherit from it.

Related to the name of the metadata section, I can modify it to coincide with name of the class, maybe is easier to read. But `meta.LoadConfigFromFile(BASIC_TRESTMETADATA_RML, "metadata");` is telling where has to look for the parameters to initialize `metadataTestClass` (the XML section called "metadata") so doesn't seem to have any problem.

DavidDiezIb: Another thing is if the test is useful or not. This was my fist try with tests and it worked, so far so good. Of couse more useful tests could be done.

jgalan: Right, if it is a fake, I agree it might be better to do not follow the naming conventions.

On the other hand, the natural behaviour should be that xml entry and c++ metadata class got the same name.

I thought `TRestMyMetadata::LoadFromConfigFile( filename, "myName")` would retrieve a RML definition as follows:

`<TRestMyMetadata name="myName" />` from `filename`, and not `<myName ... />` thats why I am worried. By the fact that this is not imposed.

jgalan: > This was introduced by @DavidDiezIb and I think its a great idea to test a class such as `TRestMetadata`.

I am not debating that

DavidDiezIb: Yes, that's right, maybe we should remove the option to select the section by hand in `LoadFromConfigFile`. I don't know if this is useful somewhere.

jgalan: Yes, I didn't know that was possible, we never use it that way, isn't @nkx?

```
Int_t TRestMetadata::LoadConfigFromFile(string cfgFileName, string sectionName) {
    fConfigFileName = cfgFileName;
    if (TRestTools::fileExists(fConfigFileName)) {
        if (sectionName == "") {
            sectionName = this->ClassName();
        }
```

So probably we should remove the second argument of that method, and just define a local variable `sectionName=this->ClassName()`?

DavidDiezIb: Exactly, I agree.

jgalan: However I am a bit surprised, I thought the second argument was the given name, so that if I got two `TRestMyMetadata` definitions I am able to select the one I am interested in:

```
<TRestMyMetadata name="md1" />

<TRestMyMetadata name="md2" />
```

So that the constructor would pick up the right definition using:
```
TRestMyMetadata *md = new TRestMyMetadata("file.rml", "md1" );
md->LoadConfigFromFile( "file.rml", "md1");
```

At least thats how I am using it inside for example `TRestAxionMagneticField` and it works.

I guess it is because either option is posible as from the definition `NameOrDeclare` at:

```
TiXmlElement* TRestMetadata::GetElementFromFile(std::string cfgFileName, std::string NameOrDecalre) 
```

But why do we need `Declare` @nkx?

lobis:
I think we should not make any more changes to this branch from this point and merge into master unless someone has any objection @rest-for-physics/core_dev @jgalan . More tests can be added in other PRs.

jgalan (approved):

juanangp (approved):

jgalan:
The upgrade at #165 should allow now to include a test using directly `TRestRun`.

Issue #133, opened 2022-02-09
Remove FindROOT.cmake


lobis:
Currently in the repository `/cmake` directory we have some `Find*.cmake` macros that in my opinion should be removed. In particular `FindROOT.cmake`.

when invoking `find_package(ROOT)` cmake macro, it will search for the corresponding macro, first in the project directory. If it doesn't find it there, it will search on some other standard directories and find it (if ROOT is installed and sourced). The problem is that since we have our own `FindROOT.cmake` macro, its overwritting the installation macro. Our macro is probably based on some specific old root version. We should use the current root installation macro.

The fix should be as simple as removing the old macro, however there are some very strange bugs happening (which took my quite a while to relate to this macro...).

In particular when doing the `make` it will fail in the first line giving:
```
/bin/sh: 1: -f: not found
```

I want to get it removed now because I think its causing issues when linking to a newer version of Garfield, throwing errors like:

```
CMake Error at cmake/MacroRootDict.cmake:509 (add_library):
  Target "RestDetector" links to target "ROOT::Geom" but the target was not
  found.  Perhaps a find_package() call is missing for an IMPORTED target, or
  an ALIAS target is missing?
Call Stack (most recent call first):
  source/libraries/detector/CMakeLists.txt:30 (COMPILELIB)


CMake Error at cmake/MacroRootDict.cmake:509 (add_library):
  Target "RestDetector" links to target "ROOT::Gdml" but the target was not
  found.  Perhaps a find_package() call is missing for an IMPORTED target, or
  an ALIAS target is missing?
Call Stack (most recent call first):
  source/libraries/detector/CMakeLists.txt:30 (COMPILELIB)


CMake Error at cmake/MacroRootDict.cmake:509 (add_library):
  Target "RestDetector" links to target "ROOT::Graf3d" but the target was not
  found.  Perhaps a find_package() call is missing for an IMPORTED target, or
  an ALIAS target is missing?
Call Stack (most recent call first):
  source/libraries/detector/CMakeLists.txt:30 (COMPILELIB)


-- Generating done
CMake Generate step failed.  Build files cannot be regenerated correctly.

[Failed to reload]
```

I have no idea what we are doing that is causing the `-f` error, any ideas are welcome.

jgalan:
I dont know, perhaps if that macro is removed there is a CMAKE variable that allows to define where cmake macros can be found?

lobis:
[...]

PR #136, opened 2022-02-14
Fix `FindROOT.cmake` in case `thisroot.sh` wasn't sourced

Vindaar:
Vindaar  [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/fixFindRoot)

Related to #133.

While helping @jovoy, we stumbled over the `FindROOT.cmake` script which currently:
- neither checks whether `ROOTSYS` is defined (i.e. whether `thisroot.*` was sourced)
- nor just uses the `root-config` to get all information it needs.

So this PR:
- throws a fatal error on Windows in case `ROOTSYS` is undefined
- uses the result of `root-config --bindir` on unix, as that is all we need to determine the path to `rootcint`.

I'm neither a cmake expert, nor sure whether it's wise in the first place to continue if `ROOTSYS` is not defined. I can't foresee if later not having sourced the file leads to even harder to debug issues. Alternatively, we could also just check for the existence of the variable on unix and fatal error in the same way.

If `FindROOT.cmake` is buried soon, just make sure the behavior of the alternative is sane & checks for this as well.

@rest-for-physics/core_dev

### DIFF
--- a/cmake/FindROOT.cmake
+++ b/cmake/FindROOT.cmake
@@ -1,15 +1,15 @@
 # - Finds ROOT instalation
-# This module sets up ROOT information 
+# This module sets up ROOT information
 # It defines:
 # ROOT_FOUND          If the ROOT is found
 # ROOT_INCLUDE_DIR    PATH to the include directory (deprecated)
 # ROOT_INCLUDE_DIRS   PATH to the include directory
 # ROOT_LIBRARIES      Most common libraries
-# ROOT_LIBRARY_DIR    PATH to the library directory 
+# ROOT_LIBRARY_DIR    PATH to the library directory
 # ROOT_BIN_DIR	      PATH to the excutables directory
 # ROOT_PYTHONVER      Compatible python version string
 
-# First search for ROOTConfig.cmake on the path defined via user setting 
+# First search for ROOTConfig.cmake on the path defined via user setting
 # ROOT_DIR
 
 if (CMAKE_SYSTEM_NAME MATCHES "Windows")
@@ -19,7 +19,7 @@ if (CMAKE_SYSTEM_NAME MATCHES "Windows")
   endif()
 
   if(NOT ROOTSYS)
-	set(ROOT_FOUND FALSE)
+    set(ROOT_FOUND FALSE)
   else()
     set(ROOT_FOUND TRUE)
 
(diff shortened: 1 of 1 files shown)


lobis:
Thanks for this.

It is my goal to remove the `FindROOT.cmake` macro soon (in https://github.com/rest-for-physics/framework/pull/111).

Now the way to find root is to call the standard cmake macro: `find_package(ROOT REQUIRED)` which will throw a standard cmake exception when ROOT is not found, so it is not necessary to check for `ROOTSYS` since it will always be defined if the `find_package` macro doesn't throw.

Regarding `rootcint` in the mentioend PR it is defined as `set(ROOTCINT_EXECUTABLE $ENV{ROOTSYS}/bin/rootcint)` which should always work if run after `find_package`. I have now read https://root.cern/manual/integrate_root_into_my_cmake_project/ more carefully and have defined in a better way in 891522cea472a27fca81f4e9f338dbeb0109277f. This should work in all platforms but I cannot test it in Windows or macOS.

jgalan (approved):
Pipeline is green so for me is good to merge and solve installation problems.

lobis (approved):
FindROOT.cmake macro will be deprecated soon.

jgalan:
I guess it is fine to merge now. @nkx is using Windows, but I believe he will not get in troubles with this patch?

Vindaar:
> Thanks for this.
> 
> It is my goal to remove the `FindROOT.cmake` macro soon (in #111).
> 
> Now the way to find root is to call the standard cmake macro: `find_package(ROOT REQUIRED)` which will throw a standard cmake exception when ROOT is not found, so it is not necessary to check for `ROOTSYS` since it will always be defined if the `find_package` macro doesn't throw.

I don't quite understand how that will work. If I understand the cmake docs correctly:

https://cmake.org/cmake/help/latest/command/find_package.html

it says that `find_package(foo, ...)` simply dispatches to a `FindFoo.cmake` file. So if one removes the `FindROOT.cmake` usage of `find_package` breaks, no?

> I guess it is fine to merge now. @nkx is using Windows, but I believe he will not get in troubles with this patch?

Unless I made some mistake, nothing should change in case the ROOT installation is set up correctly. If it wasn't there probably would have been other errors for @nkx111.

lobis:
> it says that `find_package(foo, ...)` simply dispatches to a `FindFoo.cmake` file. So if one removes the `FindROOT.cmake` usage of `find_package` breaks, no?

@Vindaar 
No, `find_package` will try to find this package looking first into standard places. If root is sourced correctly it will find `ROOTConfig.cmake` (or something similar) probably in `$ROOTSYS/cmake` and load it. I removed the `FindROOT.cmake` because it is shadowing this functionality, we're using some frozen version of `ROOTConfig.cmake` (named `FindROOT.cmake`) instead of the current version packaged in the installation directory.

nkx111:
> Unless I made some mistake, nothing should change in case the ROOT installation is set up correctly. If it wasn't there probably would have been other errors for @nkx111.

No errors for me. Actually on Windows I only do the coding. When building REST I always switch to linux.

nkx111 (approved):

PR #143, opened 2022-02-22
TRestMetadata::InitFromConfigFile is now implemented

jgalan:
jgalan  [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/generic_optics)

The first use case will be at rest-for-physics/axionlib#13 thats why the branch is named `generic_optics`.

List of changes:
- `TRestMetadata::InitFromConfigFile` has now been implemented to automatically read basic parameters.
- Images at the documentation have been removed from framework repository. They will be moved to each dedicated library.
- `REST.conf` was updated to include the libraries images path.


nkx111:
But then TRestMetadata would not be an abstract class anymore, with so abstract name it has.

jgalan:
Yes, it can be kept abstract by protecting the constructors. Thats what I did also at the commit

nkx111 (approved):

nkx111:
We can remove the repeated code in `TRestEventProcess::InitFromConfigFile()`

jgalan:
There is a difference between both, that `TRestEventProcess` has access to `fRunInfo`, and processes are able to use [TRestMdClass::fMember]

PR #147, opened 2022-03-09
Updated CMake for inclusion of REST via find_package

lobis:
lobis  [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/lobis-patch-cmake_find_package)

Changes to the CMake files to support inclusion of REST via `find_package(REST)`.

Added new command: `rest-config --prefix` which returns the installation path of REST (doesn't have to be the same as the current REST_PATH).

When doing `find_package(REST)` it will check that the current REST_PATH and the installation path of REST match. It will throw an error if they don't.

### DIFF
--- a/cmake/scripts/RESTConfig.cmake
+++ b/cmake/scripts/RESTConfig.cmake
@@ -2,9 +2,17 @@
 # This cmake file will enable to find REST from another CMake project via "find_package(REST)"
 # It will define useful variables such as "REST_PATH" or "REST_LIBRARIES" which makes linking easier
 
-set(REST_PATH $ENV{REST_PATH})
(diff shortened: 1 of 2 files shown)


lobis:
relevant PR: https://github.com/rest-for-physics/framework/pull/147

jgalan (approved):
Ok, but how could it happen that `rest-config --prefix` is not the same as `REST_PATH`?

lobis:
> Ok, but how could it happen that `rest-config --prefix` is not the same as `REST_PATH`?

The user could have changed the `REST_PATH` env variable after sourcing REST, `rest-config --prefix` is safer in general, in a similar way to `root-config --prefix` vs `$ROOTSYS`.

juanangp (approved):

PR #149, opened 2022-03-10
Fix installation issue due to Garfield

lobis:
lobis  [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/lobis-patch-garfield-install-naf-iaxo)

Fix installation on NAF-IAXO.

I renamed `FindGarfield.cmake` to `FindGarfieldOld.cmake` so that it is not found by `find_package`. It can only be called from the "old way" (as a cmake script) via `include(FindGarfieldOld)`.

Related to https://github.com/rest-for-physics/detectorlib/pull/35


lobis:
The problem has been identified as an issue with the Garfield++ code, not REST code. PR https://gitlab.cern.ch/garfield/garfieldpp/-/merge_requests/285, together with this PR, should fix the issue.

PR #158, opened 2022-03-22
Garfield compilation fix

jgalan:
jgalan  [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/jgalan_garfield_fix)

CMakeLists fixes related to Garfield compilation of old versions.

Related to rest-for-physics/framework#37

### DIFF
--- a/cmake/FindGarfieldOld.cmake
+++ b/cmake/FindGarfieldOld.cmake
@@ -1,35 +1,23 @@
-
-# - Try to find Garfield++ instalation
+# - Try to find Garfield++ installation
 # This module sets up Garfield information
 # It defines:
-# Garfield_FOUND          If Garfiled++ is found
+# Garfield_FOUND          If Garfield++ is found
 # Garfield_INCLUDE_DIRS   PATH to the include directories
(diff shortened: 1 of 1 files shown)


lobis (approved):
The changes are fine (I took the opportunity to make some formatting changes and remove old comments).

However it doesn't make sense that this file suddenly stops working, there must be a problem with the sultan installation of Garfield I think. For example, at `/programas/garfield/6.24.02` there are two "includes", one `include` and one `Include`, something must be wrong.

I would reinstall the same version of Garfield and check if this PR is really needed before merging.

Or just install the latest Garfield version and work with that, and probably drop support for old Garfield altogether, which will save time on the long run I think.

jgalan:
The good one is `include`, it is the one generated by the installation. `Include` is the one from the sources. Both should have the same content anyway. This happens because the install path is defined as `GARFIELD_HOME`. I believe this is because it was like that at the original installation instructions of Garfield.

The changes happening at this PR had no effect on solving my compilation problem. The problem was solved at `detectorlib`. The reason behind is that the include directories were not the right ones.

juanangp (approved):

PR #161, opened 2022-03-24
Updating PR template

jgalan:
jgalan  [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/jgalan_pr_template)

This is a simple PR to update the PR message template that appears when creating a new PR. 

Feel free to suggest changes now, the PR template is a relevant communication channel!

On the other hand, I believe we should generate a new version framework release once each PR has been approved.

2.3.10 --> X.Y.Z

X: Major change. For example, Migrating from ROOT6 to ROOT7. or from RML to YML.
Y: Official data will start to be produced. To define a point in time where a experiment/project started to take physics data.
Z: Any PR increases this number.

Changing version is basically based on 3 steps:
1. Adding any new library releases that were generated since last framework release as a submodule,
2. TRestVersion.h generation using a script, 
3. creating a Git tag, and 
4. placing the PR changes into the release notes.

@rest-for-physics/core_dev


.github/pull_request_template.md
```diff
@@ -1,10 +1,21 @@
...
 
 Changes proposed in this pull-request:
--
--
--
+- Upgraded ...
+- Pipeline validation ...
+- Documented ...
+- Example ...
+- Script ...
```
juanangp, line 10: I think you are missing other descriptions such as bug-fix, or feature. Would be a good idea to provide an example of a PR in every case. Also, you can point out to the different labels that are available, perhaps we should review the labels as well.

jgalan: I think that is contemplated at the upper line `Fixes #`

For me the `-` are points that the one writing the PR description should fill in. So this template is a guide/reminder of which points should be written. Not meaning that all of them should.

Making it too long could be counter-productive.

About the labels, perhaps this could be defined at the contribution guide, then we review the labels on GitHub.

jgalan:
If you agree, could we approve this PR? If not, please, propose a change through a direct commit into this PR.

jgalan:
No, if the size is lower than 100, the PR size badge will show green, above 100 orange, and above 200 red.

juanangp (approved):

PR #168, opened 2022-03-27
Removed `using namespace std` from header files and fix errors

lobis:
lobis Large: 1097 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/lobis-remove-namespace-std-from-headers)

This is the smallest PR (that I could think of) that removes `using namespace std;` from all headers files and fixed all compilation errors.

It must be merged with the associated PRs from all libraries / packages which appear mentioned below.

Even though this PR should not have any effect on the framework, it may produce some unintended effects with some classes introduced by @nkx111 regarding `endl` in `TRestStringOutput.h`.

The previous implementation relied on having `endl` pointing to `std::endl` by `using namespace std` on the header. Now you cannot use `endl` on `cout` statements (because of missing namespace, unless you include it) and you cannot use `std::endl` on the `metadata << ` type statements, because of this "collision". I think this should not be the case and should be fixed.

PR that also have to be merged along this one:

* https://github.com/rest-for-physics/connectorslib/pull/9
* https://github.com/rest-for-physics/tracklib/pull/14
* https://github.com/rest-for-physics/geant4lib/pull/39
* https://github.com/rest-for-physics/detectorlib/pull/43


lobis:
For some reason pipeline https://gitlab.cern.ch/rest-for-physics/framework/-/jobs/20613842 is failing, but looks like its not using the corresponding version of the `raw` library, I think it should not fail.

Vindaar:
What's the reasoning behind the sudden change? Just the fact that REST barely uses any STL stuff and is essentially a ROOT project instead? 

Note that I do agree that pulling in `std` everywhere is rather problematic in C++.

lobis:
> What's the reasoning behind the sudden change? Just the fact that REST barely uses any STL stuff and is essentially a ROOT project instead? 
> 
> Note that I do agree that pulling in `std` everywhere is rather problematic in C++. 

It's part of our goal of cleaning up old bad practices, it's something we wanted to do for a while and now it's finally done ✅, it didn't take that much effort. There wasn't a change on design philosophy, I don't think there was a reason for it to be that way in the first place.

It makes no sense to use standard namespace in the header files even if with our current usage is not that problematic.

REST is supposed to be a framework you can integrate into other projects (or so I think, even though in practice it's not the case) and having to carry over this namespace is a big (potentially unavoidable?) problem as it could collide with some other library using the same namespace (which I guess would also be a bad practise but...).

source/framework/core/src/TRestAnalysisTree.cxx
```diff
@@ -591,11 +591,11 @@ void TRestAnalysisTree::PrintObservables() {
     }
 
     cout.precision(15);
-    std::cout << "Entry : " << fReadEntry << std::endl;
-    std::cout << "> Event ID : " << GetEventID() << std::endl;
-    std::cout << "> Event Time : " << GetTimeStamp() << std::endl;
-    std::cout << "> Event Tag : " << GetSubEventTag() << std::endl;
-    std::cout << "-----------------------------------------" << std::endl;
```
juanangp, line 598: Do you understand why we have to revert `std::endl` to `endl`

lobis: > Do you understand why we have to revert `std::endl` to `endl`

Can you rephrase?

juanangp: Does it arises from `struct endl_t` on `TRestStringOutput.h`? I think we should fix it in this PR...

lobis: I am not sure if I would complicate more this PR as its already big enough, perhaps @jgalan has a opinion. I agree that this should be fix, but perhaps on another PR if its not an easy fix. So far this doesn't seem to cause any problem, but it may have been unnoticed.

jgalan: `TRestStringOutput.h` was designed by @nkx, probably he has a better opinion than myself.

On the other hand, you need to create a new branch `lobis-remove-namespace-std-from-headers` at rawlib to fix compilation issues at this PR.

lobis: > On the other hand, you need to create a new branch `lobis-remove-namespace-std-from-headers` at rawlib to fix compilation issues at this PR.

But this branch already exists, I don't understand the error. https://github.com/rest-for-physics/rawlib/tree/lobis-remove-namespace-std-from-headers

nkx111: > The problems with endl arise from TRestStringOutput.h

What is the problem? In principle it is OK to remove `using namespace std`, because once we do that, we need to state `std::end`, which induces less ambiguity to the compiler. There will be less problems for my `endl_t`

lobis: > > The problems with endl arise from TRestStringOutput.h
> 
> What is the problem? In principle it is OK to remove `using namespace std`, because once we do that, we need to state `std::end`, which induces less ambiguity to the compiler. There will be less problems for my `endl_t`

If I remember correctly one of the problems is that you need to use `std::endl` (explicitly, event when `using namespace std` is declared) when using `std::cout` or it will throw an error. Also you need to use `endl` (without `std::`) when using `metadata <<` and such or it will throw an error. I think it shouldn't behave like this.

source/framework/core/src/TRestAnalysisPlot.cxx
```diff
@@ -550,7 +550,7 @@ Int_t TRestAnalysisPlot::GetPlotIndex(TString plotName) {
 }
 
 TRestAnalysisTree* TRestAnalysisPlot::GetTree(TString fileName) {
-    if (fRun->GetInputFile() != NULL && fRun->GetInputFile()->GetName() == fileName){
+    if (fRun->GetInputFile() != NULL && fRun->GetInputFile()->GetName() == fileName) {
```
juanangp, line 553: `nullptr` or `if( (fRun->GetInputFile()) && (fRun->GetInputFile()->GetName() == fileName))`

lobis:
I think this PR can be merged and causes no issues. It would be nice to fix the `endl` thing but I think it doesn't cause problems, its just annoying.

Since this is a big PR I think it should be merged as soon as possible before I need to fix too many conflicts. Let me know if you want something changed @jgalan @juanangp @nkx111

lobis:
Looks like the pipeline is failing because of a bug on the `pull-submodules.py` script or similar, its not picking the correct version of `rawlib`.

juanangp:
Can you post in the description the related PR for the other libraries and/or packages?
Still cannot see any `lobis-remove-namespace-std-from-headers` in `rawlib` https://github.com/rest-for-physics/rawlib/branches

jgalan:
> Can you post in the description the related PR for the other libraries and/or packages?
> Still cannot see any `lobis-remove-namespace-std-from-headers` in `rawlib` https://github.com/rest-for-physics/rawlib/branches

Thats probably why `pull-submodules.py` does not pick up the right commit?

juanangp:
> Thats probably why `pull-submodules.py` does not pick up the right `rawlib` commit?

I don't think so, looks like `lobis-remove-namespace-std-from-headers`  `rawlib` branch is already merged in `master`. But there are still compilations errors. I checked locally and doesn't compile. I guess further modifications are needed after removing global `using namespace std` from framework.

lobis:
> Can you post in the description the related PR for the other libraries and/or packages? Still cannot see any `lobis-remove-namespace-std-from-headers` in `rawlib` https://github.com/rest-for-physics/rawlib/branches

It is merged into master

lobis:
> > Thats probably why `pull-submodules.py` does not pick up the right `rawlib` commit?
> 
> I don't think so, looks like `lobis-remove-namespace-std-from-headers` `rawlib` branch is already merged in `master`. But there are still compilations errors. I checked locally and doesn't compile. I guess further modifications are needed after removing global `using namespace std` from framework.

If you look into the pipeline, you will see its not pulling the correct branch of rawlib, as the code doesn't match. @jgalan and I already looked into it and found no solution or reason why it does this, but it compiles locally

lobis:
> Screenshot 2022-03-31 at 14 26 03

It should be pulling https://github.com/rest-for-physics/rawlib/commit/d08c8fa0164066d15fb5939858f79b7ad5cbedbe which is the latest commit from `master` branch at the moment, right?

jgalan:
Thats when I do `git log` inside `rawlib`

```
commit 9eff003891217e35acc107bf3d01efee1c2e6f37 (HEAD)
Merge: 05b82eb d08c8fa
Author: Javier Galan <javier.galan@unizar.es>
Date:   Thu Mar 31 14:32:03 2022 +0200

    Merge branch 'master' of github.com:rest-for-physics/rawlib into HEAD
```

jgalan:
I believe it is ok, it is pulling the most updated from `master` it is just that the commit number is not the one we expected.

juanangp:
This issue could be because we just checkout the latest branch that passed the CI?

What does `CI_COMMIT_BRANCH` stands for?
```
python3 pull-submodules.py --force --dontask --latest:${CI_COMMIT_BRANCH}
```

lobis:
> This issue could be because we just checkout the latest branch that passed the CI?
> 
> What does `CI_COMMIT_BRANCH` stands for?
> 
> ```
> python3 pull-submodules.py --force --dontask --latest:${CI_COMMIT_BRANCH}
> ```

If it pulled master in absense of the `${CI_COMMIT_BRANCH}` it would work. But since it pulls the fixed submodule version of the framework it fails, perhaps we should change the default behaviour or add this as an option.

juanangp (approved):

lobis:
can we approve this merge request @jgalan @rest-for-physics/core_dev ? It doesn't pass the pipeline because submodules need to be updated, but it should be OK. what shall we do?

Also I did a very small change to the new `curl` in [45ca1db](https://github.com/rest-for-physics/framework/pull/168/commits/45ca1db3a97f0a49881f4d2f36709bce226a6f0e), I hope its ok. I think it would be nice to have a unit test in place for new additions such as this one, in order to promote testing.

jgalan:
Fixed the issue with the process validation, now it seems the problem is somewhere else during compilation.

lobis:
> cdacf67

the problem is due to `restG4`, its not using the `master` branch for the compilation

jgalan:
In principle it says it is pulling master 

```
packages/restG4 --> Pulling branch : master  [[92m OK [0m] (cc7486c)
```

Why don't you create a `lobis-remove-namespace-std-from-headers` at restG4 to try out?

jgalan:
You could also try at the pipeline something like:

```
cd source/packages/restG4
git log
```

lobis:
> In principle it says it is pulling master
> 
> ```
> packages/restG4 --> Pulling branch : master  [�[92m OK �[0m] (cc7486c)
> ```
> 
> Why don't you create a `lobis-remove-namespace-std-from-headers` at restG4 to try out?

but the correct commit is `[49d0345](https://github.com/rest-for-physics/restG4/commit/49d0345b2705d8e884d4cf8565688c303fb78977)`. I am sure its not using the correct version because of the logs, its calling different functions, for example [49d0345](https://github.com/rest-for-physics/restG4/commit/49d0345b2705d8e884d4cf8565688c303fb78977). I could fix it creating a dummy branch but we should find the root cause

jgalan:
Yes, I guess creating the dummy branch will at least help to constrain and identify the problem.

jgalan:
Up to now there was not problems, perhaps it happens when the last commit is a merge commit. I usually keep updated with master anytime, and then this problem did not appear before.

Solution might be modifying the way we pull submodules at `pull-submodules.py`

```
                         if latest == 1:
                             command = 'git ls-remote --heads ' + url + ' ' + frameworkBranchName + ' | wc -l'
                             branchExistsPcs = subprocess.run(command, shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)

                             branchToPull = "master"
                             if( branchExistsPcs.stdout.decode("utf-8").rstrip("\n") != "0"):
                                 branchToPull = frameworkBranchName
                                 print( " --> Pulling branch : " + branchToPull + "  " , end='' )

                             p = subprocess.run('cd {}/{} && git pull --tags origin {}'.format(root, submodule, branchToPull), shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
                             if(debug):
                                 print (p.stdout.decode("utf-8"))
                                 print (p.stderr.decode("utf-8"))
                             errorOutput = p.stderr.decode("utf-8")
                             if errorOutput.find("failed") != -1 or errorOutput.find("error") != -1:
                                 print ("[\033[91m Failed \x1b[0m]")
                                 if(debug):
                                     print ("Message: ")
                                     print (errorOutput)
                                 continue
```

lobis:
@jgalan @juanangp  I updated the script with https://github.com/rest-for-physics/framework/pull/168/commits/764f6286040735f66821252ac1f47a75003cf8c9 and now the pipeline passes, its now pulling master branch when the named branch is not found. The only behavioural change is this commit, all other changes have no effect on usage of the script.

Please let me know if you think this can be merged now.

nkx111 (approved):

jgalan (approved):

PR #173, opened 2022-03-31
Using `cpr` as http library

lobis:
lobis  [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/lobis-jgalan_tools_post)

I added this library via cmake to handle `http` requests so we don't need to force the user to install external dependencies.

It looks like there is some problem on the cmake however, I think due to the `COMPILELIB()` macro...


jgalan:
I believe this should be an issue and not a PR, if not the list of PR is getting too long.

PR #174, opened 2022-04-01
Made some methods const

lobis:
lobis Large Medium [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/lobis-add-const-to-getters)

Changes required for https://github.com/rest-for-physics/geant4lib/pull/41.

Some methods are made `const`, usage remains the same.

I also noticed abug in [e6da8d7](https://github.com/rest-for-physics/framework/pull/174/commits/e6da8d7072b83819ff977095ab57b6f4e1490231), perhaps the same bug is present elsewhere.


lobis:
pipeline seems to fail giving an odd error with observables on the tree... I will look into it, no idea what can be, anyone has seen this before @rest-for-physics/core_dev ?

jgalan:
I have seen that before, because if something affects the observables previously registered at some point in time with the ones obtained now, then the pipeline will not succeed. However, there is no clue I could give since the reasons could be completely unconnected.

source/framework/core/src/TRestEvent.cxx
```diff
@@ -73,26 +73,25 @@ void TRestEvent::Initialize() {
...
     fPad = nullptr;
 }
 
 //////////////////////////////////////////////////////////////////////////
 /// Set the time of the event
 ///
 void TRestEvent::SetTime(Double_t time) {
-    Int_t sec = (Int_t)time;
-    Int_t nsec = (Int_t)((time - sec) * 1E9);
+    auto seconds = (Int_t)time;
```
juanangp, line 83: Perhaps thiis is the issue? I don't know if `auto` is capable to assign `Int_t` but should be equivalent to `int`. Unless is doing something weird...

source/framework/core/src/TRestHits.cxx
```diff
@@ -842,20 +851,19 @@ Double_t TRestHits::GetDistanceToNode(Int_t n) {
     return distance;
 }
 
-Int_t TRestHits::GetMostEnergeticHitInRange(Int_t n, Int_t m) {
-    Int_t maxEn = 0;
+Int_t TRestHits::GetMostEnergeticHitInRange(Int_t n, Int_t m) const {
+    double maxEnergy = 0;
```
juanangp, line 855: This can also change the pipeline behaviour but it is indeed a bug, since energy should be `double`

juanangp:
Looking at the pipeline there is something wrong with the hits (left failure; right suceed) 

image

source/framework/core/src/TRestHits.cxx
lobis: yes, this is the only thing I can think of... but I don't think this method is used right? hopefully its not used anywhere since it was bugged.

lobis:
> Looking at the pipeline there is something wrong with the hits (left failure; right suceed)
> 
> image

I added back the bug, lets see if it gets fixed.

source/framework/core/src/TRestHits.cxx
```diff
@@ -852,7 +852,7 @@ Double_t TRestHits::GetDistanceToNode(Int_t n) const {
 }
 
 Int_t TRestHits::GetMostEnergeticHitInRange(Int_t n, Int_t m) const {
-    double maxEnergy = 0;
+    Int_t maxEnergy = 0;
```
jgalan, line 855: ups

jgalan: A commit like that should be in a stand-alone PR!

jgalan: Not in a "made some methods const"

lobis: Well adding it back doesn't fix it. I think this bug or similar might be in more methods of the `TRestHits` class. Any ideas to fix the pipeline failure?

source/framework/core/inc/TRestEvent.h
```diff
@@ -105,6 +105,6 @@ class TRestEvent : public TObject {
     // Destructor
     virtual ~TRestEvent();
 
-    ClassDef(TRestEvent, 1);  // REST event superclass
+    ClassDef(TRestEvent, 2);  // REST event superclass
```
jgalan, line 108: No need to change the ClassDef version if metadata members have not been changed.

source/framework/core/src/TRestHits.cxx
jgalan: The issue could also be at another library that was merged to master recently.

source/framework/core/inc/TRestEvent.h
lobis: OK, I am never sure of this.

source/framework/core/src/TRestHits.cxx
lobis: so I think we can merge this into master right?

juanangp: Well, I think we should not merge unless we understand the issue. Or at least if the issue have been originated in this or related PR.

The pipeline for current master is passing?

juanangp:
I think I found something odd in rest-for-physics/connectorslib#10 perhaps related with the pipeline failure...

juanangp:
> I made some fixes but still pipeline fails...

There is something nasty  because the number of processed events is zero, which means that the error is different from before. Either a bug has been introduced with the new changes or elsewhere.

juanangp:
Just checkout this branch and I realize that the pipeline is extremelly slow. Check below what I get while comparing with an older branch:

image

Doesn't seems related to a particular process, but all processes are slow. Do you know something that has been recently introduced which can explain this time difference? I guess should be under framework...

On the other hand, it would be good to check the running time in the pipeline validation to catch these kind of issues.

jgalan:
> Just checkout this branch and I realize that the pipeline is extremelly slow. Check below what I get while comparing with an older branch:
> 
> image
> 
> Doesn't seems related to a particular process, but all processes are slow. Do you know something that has been recently introduced which can explain this time difference? I guess should be under framework...
> 
> On the other hand, it would be good to check the running time in the pipeline validation to catch these kind of issues.

Test should be done in the same local machine. It could happen that the load from the server at a given time would affect, i imagine.

source/framework/core/src/TRestHits.cxx
jgalan:
```
jgalan@sultan:~/rest-framework/source$ grep -rnw . -e "GetMostEnergeticHitInRange"
./libraries/track/src/TRestTrackBlobAnalysisProcess.cxx:215:        Int_t hit1 = hits->GetMostEnergeticHitInRange(0, nCheck);
./libraries/track/src/TRestTrackBlobAnalysisProcess.cxx:216:        Int_t hit2 = hits->GetMostEnergeticHitInRange(nHits - nCheck, nHits);
```

jgalan:
At the end what it was making the pipeline fail? 

Coming back to the processing time. Perhaps we should add as a validation pipeline that the processing time of a reference processing chain does not exceed a certain value. But not sure if it can be assured that it will be comparable at each runner.

juanangp:
> Test should be done in the same local machine. It could happen that the load from the server at a given time would affect, i imagine.

I am running in sultan, so I think that the issue is isolated to the code and not to the server.

juanangp:
> At the end what it was making the pipeline fail?

Aparently, we introduced another bug in https://github.com/rest-for-physics/connectorslib/pull/10, which have to be fixed.
 
> Coming back to the processing time. Perhaps we should add as a validation pipeline that the processing time of a reference processing chain does not exceed a certain value. But not sure if it can be assured that it will be comparable at each runner.

Yes, I was thinking of checking the cpu time, that I think is what the image that I posted before is doing. Can this info be added as metadata or so?

jgalan:
> Yes, I was thinking of checking the cpu time, that I think is what the image that I posted before is doing. Can this info be added as metadata or so?

I imagine we could add a new metadata member to `TRestEventProcess::fAverageProcessingTime` so that every process registers the time required for its own execution.

lobis:
> At the end what it was making the pipeline fail?

I have no idea, I made some changes to the pipeline scripts to have better debuging and suddenly it worked.

pipeline/ValidateTrees.C
```diff
@@ -1,52 +1,60 @@
...
-
-    if (tr->GetEntries() != tV->GetEntries()) {
+
+Int_t ValidateTrees(const char* validationFilename, const char* inputFilename = "results.root") {
+    TFile inputFile(inputFilename);
+    TTree* inputTree = inputFile.Get<TTree>("bTree");
+    std::vector<double>* inputObservableVector = 0;
+    inputTree->SetBranchAddress("doubleObservables", &inputObservableVector);
+
+    TFile validationFile(inputFilename);
```
juanangp, line 13: I believe this should be `TFile validationFile(validationFilename);`

lobis: > I believe this should be `TFile validationFile(validationFilename);`

good catch thanks, its now fixed.

lobis:
Looks like [53d73ec](https://github.com/rest-for-physics/connectorslib/pull/10/commits/53d73ec8fa2cff71a8f71783fc2b3cc22a559fea) finally fixed the pipeline problems, thanks @juanangp.

So regarding that, this and all related PR can be merged.

However the problem with the speed... I can't think how this PR has anything to do with that, specially since it happens on all kinds of analysis proecsses, not only those that would be affected by this PR, what do you think about merging @jgalan @juanangp ?

lobis:
@juanangp and I have checked together if these changes produce any performance change and haven't been able to reproduce, we made a clean install using these changes `lobis-add-const-to-getters` on all corresponding branches. Our conclusion is that this PR does not introduce any performance overhead so it can be merged.

juanangp (approved):

DavidDiezIb (approved):

PR #175, opened 2022-04-04
Few TRestTools/TRestPhysics/TRestStringHelper upgrades motivated by TRestAxionOpticsMirror

jgalan:
jgalan Large: 299 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/jgalan_mirror)

Few tools added and/or reviewed which are required by recent implementation at rest-for-physics/axionlib#22

- `TRestTools::ExportBinaryTable(fname,data)` has been added. It will write a binary file with name `fname` with the contents of `std::vector<std::vector <double/float/int>> &data`.

- `TRestTools::GetMin/MaxValueFromTable` now operates over the full table if no column is specified.

- `TRestStringHelper::StringToElements` prototyping has been simplified, and method is now overloaded. Usage of new implemented method `StringToElements( string in, "[", ",", "]" )` where the method will find the first string coincidence of "[" and extract the elements from the given format.

- `TRestStringHelper::DoubleToString` it takes a double number and returns it as a string.

- `TRestPhysics::GetConeVectorIntersection` added to find the intersection of a straight with a cone.

- `TRestTools::GetFileNameExtension` added to facilitate retrieving the extension of a filename.

- `TRestTools::IsBinaryFile` and `TRestTools::GetBinaryFileColumns` methods added.

- `TRestTools::ReadBinaryTable` will attempt to read the number of columns from `N000i/f/d` filename extension format.

@rest-for-physics/core_dev


source/framework/tools/src/TRestTools.cxx
```diff
@@ -212,48 +239,72 @@ template int TRestTools::ReadBinaryTable<Double_t>(string fName, std::vector<std
...
 /// This method is available for tables of type Float_t, Double_t and Int_t.
 ///
 template <typename T>
-T TRestTools::GetMinValueFromTable(std::vector<std::vector<T>> data, Int_t column) {
-    if (data.size() == 0 || data[0].size() <= column) return 0;
-    T minValue = data[0][column];
-    for (int n = 0; n < data.size(); n++)
-        if (minValue > data[n][column]) minValue = data[n][column];
+T TRestTools::GetMinValueFromTable(const std::vector<std::vector<T>>& data, Int_t column) {
+    if (data.size() == 0) return 0;
```
lobis, line 284:
```suggestion
    if (data.empty()) return 0;
```

In general we should use `std::vector::empty()` to check for empty vectors instead.

lobis (approved):

source/framework/tools/src/TRestTools.cxx
jgalan: why?

source/framework/tools/src/TRestTools.cxx
```diff
@@ -211,50 +248,118 @@ template int TRestTools::ReadBinaryTable<Float_t>(string fName, std::vector<std:
...
-T TRestTools::GetMaxValueFromTable(std::vector<std::vector<T>> data, Int_t column) {
-    if (data.size() == 0 || data[0].size() <= column) return 0;
-    T maxValue = data[0][column];
-    for (int n = 0; n < data.size(); n++)
-        if (maxValue < data[n][column]) maxValue = data[n][column];
+T TRestTools::GetMaxValueFromTable(const std::vector<std::vector<T>>& data, Int_t column) {
+    if (data.size() == 0) return 0;
+    if (column != -1 && data[0].size() <= column) return 0;
+
+    T maxValue = data[0][0];
```
juanangp, line 307: This migth be wrong, what happen if `data[0][0]` is bigger than any element on `data[0][colum]` ?

source/framework/tools/src/TRestTools.cxx
```diff
@@ -211,50 +248,118 @@ template int TRestTools::ReadBinaryTable<Float_t>(string fName, std::vector<std:
...
 /// This method is available for tables of type Float_t, Double_t and Int_t.
 ///
 template <typename T>
-T TRestTools::GetMaxValueFromTable(std::vector<std::vector<T>> data, Int_t column) {
-    if (data.size() == 0 || data[0].size() <= column) return 0;
-    T maxValue = data[0][column];
-    for (int n = 0; n < data.size(); n++)
-        if (maxValue < data[n][column]) maxValue = data[n][column];
+T TRestTools::GetMaxValueFromTable(const std::vector<std::vector<T>>& data, Int_t column) {
+    if (data.size() == 0) return 0;
```
juanangp, line 304: I think that `if (data.empty()) return 0;` is more readable

source/framework/tools/src/TRestTools.cxx
juanangp: > This migth be wrong, what happen if `data[0][0]` is bigger than any element on `data[0][column]` ?

I see, this is addressed below

source/framework/tools/src/TRestTools.cxx
juanangp: https://stackoverflow.com/questions/743197/size-vs-empty-in-vector-why-empty-is-preferred

juanangp (approved):

PR #176, opened 2022-04-04
Encapsulating the use of CURL libraries

jgalan:
We recently included dependency with curl libraries.

Inside this PR I am encapsulating the methods using CURL so that if no curl libraries have been installed in the system the CURL functions used in the code are bypassed.

Perhaps cmake gurus prefer a different add-on. In that case, please do not hesitate to commit any changes here.

@rest-for-physics/core_dev

### DIFF
--- a/source/framework/tools/src/TRestTools.cxx
+++ b/source/framework/tools/src/TRestTools.cxx
@@ -839,6 +839,8 @@ int TRestTools::DownloadRemoteFile(string remoteFile, string localFile) {
 /// by argument, and returns the result as a string.
 ///
 std::string TRestTools::POSTRequest(const std::string& url, const std::map<std::string, std::string>& keys) {
+    std::string file_content = "";
+#ifdef USE_Curl
     CURL* curl;
     CURLcode res;
 
@@ -878,8 +880,13 @@ std::string TRestTools::POSTRequest(const std::string& url, const std::map<std::
     fclose(f);
     curl_global_cleanup();
 
-    std::string file_content = "";
     std::getline(std::ifstream(filename), file_content, '\0');
+#else
+    ferr << "TRestTools::POSTRequest. REST framework was compiled without CURL support" << endl;
+    ferr << "Please recompile REST after installing curl development libraries." << endl;
+    ferr << "Depending on your system this might be: curl-dev, curl-devel or libcurl-openssl-dev. " << endl;
+    ferr << "No file will be downloaded" << endl;
+#endif
 
     return file_content;
 }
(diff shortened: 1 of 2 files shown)


lobis (changes requested):
Looks good to me, but I think all this logic should be on the `framework` library directory, not on the root directory, even though it does the same thing this is only because we add this subdirectory first. I would do something like what I suggested [here](https://github.com/rest-for-physics/framework/pull/168/commits/45ca1db3a97f0a49881f4d2f36709bce226a6f0e), if you agree.

jgalan:
Plz @lobis commit to apply the changes as you propose directly to this PR.

lobis:
> Plz @lobis commit to apply the changes as you propose directly to this PR.

Now I think it should work, previously it was trying to add the library even if it didn't find it. It would be nice to test it on a system without this library before merging though.

jgalan:
I tested my changing the name to a library I dont have installed, and it was working

jgalan:
Perhaps at NAF-IAXO is not yet installed, although I requested it. It is `latest` being compiled nightly?

jgalan:
For me at NAF-IAXO at commit dfd08b2 it works properly, perhaps the problem was introduced later on

lobis:
> I tested by changing the name to a library I dont have installed, and it was working

oh okay, I was just using the documentation as reference, the variable when not defined is not `NOTFOUND` but `<VAR>-NOTFOUND`, perhaps this changes between versions.

lobis (approved):

source/framework/tools/src/TRestTools.cxx
```diff
@@ -878,8 +880,13 @@ std::string TRestTools::POSTRequest(const std::string& url, const std::map<std::
     fclose(f);
     curl_global_cleanup();
 
-    std::string file_content = "";
     std::getline(std::ifstream(filename), file_content, '\0');
+#else
+    ferr << "TRestTools::POSTRequest. REST framework was compiled without CURL support" << endl;
```
juanangp, line 885: I think we should create a `RESTException` for these cases. Shall I create an issue about this?

jgalan: Yes, if you like, please, explain at the issue how do you want to handle it. For the moment we use no expections, as you see, as it is it just prompts a warning message, then does nothing. Not sure if in this particular case it requires of error handling.

jgalan:
> > I tested by changing the name to a library I dont have installed, and it was working
> 
> oh okay, I was just using the documentation as reference, the variable when not defined is not `NOTFOUND` but `<VAR>-NOTFOUND`, perhaps this changes between versions.

Yes, in my case it was also the format `<VAR>-NOTFOUND` I thought the cmake instruction `MATCH` just checks the string contains the keyword, while `STREQUAL` requires is exactly the same.

source/framework/tools/src/TRestTools.cxx
juanangp: I would say wherever we have `exit()` in the code we should replace by a `throw` with an exception handler. Perhaps not for this particular case...

PR #179, opened 2022-04-06
Removing override, looks problematic with root

juanangp:
After new changes in the code, it seems that root doesn't like `override` method, see below:

image

Fix this error by removing `override` from `TRestVolumeHits`


lobis (approved):

jgalan (approved):

nkx111:
You need to use ClassDefOverride, then it will not report this warning.

PR #184, opened 2022-04-21
Few TRestTools upgrades motivated by TRestAxionSolarFlux implementations

jgalan:
jgalan Ok: 73 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/jgalan_solar_flux)

Adds few methods that are needed at rest-for-physics/axionlib new implementations

- `TRestTools::GetFileNameRoot` to get the filename without extension and without path.

- `TRestTools::TransposeTable` it changes rows by columns in a `std::vector<std::vector` table.

- `TRestMetadata` will now search for the RML in case it is not found at the path we have given in the constructor. For example, if now I do `TRestAxionSolarFlux("fluxes.rml")` it will get the `fluxes.rml` file found at the data path. This is handy so that input data and definitions are found in a single place, so there is no need to have the RML copied somewhere else.
https://github.com/rest-for-physics/axionlib-data/tree/master/solarFlux


source/framework/tools/src/TRestTools.cxx
```diff
@@ -606,6 +626,22 @@ string TRestTools::GetFileNameExtension(string fullname) {
...
 }
 
+///////////////////////////////////////////////
+/// \brief Gets the filename root as the substring found before the lastest "."
+///
+/// Input: "/home/jgalan/abc.txt" Output: "abc"
+///
+string TRestTools::GetFileNameRoot(string fullname) {
+    int pos1 = fullname.find_last_of('/', -1);
+    int pos2 = fullname.find_last_of('.', -1);
```
juanangp, line 636: Please use appropriate data types, in this case `size_t`, I have created a suggestion via PR about this https://github.com/rest-for-physics/framework/pull/197

jgalan: I see, but why you dont just make a change proposal at this PR?

source/framework/tools/src/TRestTools.cxx
```diff
@@ -606,6 +651,22 @@ string TRestTools::GetFileNameExtension(string fullname) {
...
```
juanangp, line 661:
```suggestion
    size_t pos1 = fullname.find_last_of('/', -1);
    size_t pos2 = fullname.find_last_of('.', -1);
```

source/framework/tools/src/TRestTools.cxx
juanangp: OK, I though, this was the way to create it, now I have added the suggestion, ignore this PR

source/framework/tools/src/TRestTools.cxx
jgalan: Perfect!

juanangp (approved):

nkx111 (approved):

Issue #162, opened 2022-03-24
Increasing minimum CMake version causes problems on install


lobis:
When working on https://github.com/rest-for-physics/framework/pull/129 I noticed that the pipeline was not passing due to the installation directory missing some files (the binaries and libraries). After much trial and error, I found that the problem was that I had increased the minimum required version of CMake. After reverting this change, it was fixed https://github.com/rest-for-physics/framework/pull/129/commits/848f2da8d77c520512e697478f14e6b67b00c4c9.

However since https://github.com/rest-for-physics/framework/pull/129 is using some new features of CMake (and perhaps other code is already doing the same), having `3.5` as the minimum CMake version (which is from 6 years ago), will probably lead to problems on systems that use a cmake version closer to 3.5.

I vote to bump the minimum CMake version to `3.16`, but in order to do that we need to fix this issue.

juanangp:
This issue seems related to a `cmake` bug, which was introduced in version `3.14` and seems resolved after version `3.21`, this last point I didn't check properly. Seems related with cmake policy `CMP0082` https://cmake.org/cmake/help/latest/policy/CMP0082.html and it is weird since it depends on the `cmake_minimum_version` required and not the cmake version which is used to compile the code.
For more details on this cmake bug check https://gitlab.kitware.com/cmake/cmake/-/issues/22234

There are two solutions to fix this issue:
- Use older `CMP0082` policy, that should be added at the beggining of the `CMakeList.txt`:
```
if(${CMAKE_VERSION} VERSION_LESS "3.21") 
    cmake_policy(SET CMP0082 OLD)
endif()
```
- The other solution is to place `add_subdirectory` command after any `install` command in the `CMakeList.txt`

Let me know how do you want to proceed.

PR #185, opened 2022-04-21
Added preprocessor definition via cmake STRING

lobis:
lobis Ok: 7 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/lobis-regex-lib-compile-def)

Related to https://rest-forum.unizar.es/t/detectorlib-implementation-problem-inside-the-rawlib-process/535/5

Adding definitions via CMake which appear only when corresponding libraries are selected to compile. This is useful when we want to have code in one library that uses another library that may or may not be available.

I tried to look for where this was being defined previously in the code but didn't find it (for any library) I guess it was removed at some point and never cause problems.


jgalan:
Ok, it seems to be working nice.

The only thing is, how the DEBUG statement works? Perhaps its behaviour is related to the cmake version?

```
Empty submodule dir: packages/restSQL. Option: REST_SQL=OFF
-- making build files in /home/jgalan/rest-framework/source/libraries/axion, dependency: 
-- 20 source files in total, 20 classes to generete, 0 additional source files
DEBUGAdding compile definition: REST_AxionLib
-- making build files in /home/jgalan/rest-framework/source/libraries/detector, dependency: 
-- 37 source files in total, 37 classes to generete, 0 additional source files
DEBUGAdding compile definition: REST_DetectorLib
```

jgalan (dismissed):

lobis:
> Ok, it seems to be working nice.
> 
> The only thing is the output is a bit distorted, how the DEBUG statement works? Perhaps its behaviour is related to the cmake version?
> 
> ```
> Empty submodule dir: packages/restSQL. Option: REST_SQL=OFF
> -- making build files in /home/jgalan/rest-framework/source/libraries/axion, dependency: 
> -- 20 source files in total, 20 classes to generete, 0 additional source files
> DEBUGAdding compile definition: REST_AxionLib
> -- making build files in /home/jgalan/rest-framework/source/libraries/detector, dependency: 
> -- 37 source files in total, 37 classes to generete, 0 additional source files
> DEBUGAdding compile definition: REST_DetectorLib
> ```

https://cmake.org/cmake/help/latest/command/message.html

In my cmake version it works fine, I guess this did not exist in your cmake version and thats why it concatenates the strings. DEBUG messages are intended for developers and not users, in my opinion they are a middle ground between STATUS messages and comments, you can change the DEBUG to STATUS to display the mesage easier than to change it from a comment to a message.

There are many small problems similar to this that can happen due to allowing users to use very old cmake versions, we should change the minimum cmake version but its not possible yet because of https://github.com/rest-for-physics/framework/issues/162, I couldn't fix it. I encourage anyone @rest-for-physics/core_dev to try to fix this issue as I think its a bit critical.

juanangp (approved):

jgalan:
I woudnt mind that the DEBUG statement is in STATUS statement. Anyway a non-developer user will not look most of the information coming out from cmake.

lobis:
> I woudnt mind that the DEBUG statement is in STATUS statement. Anyway a non-developer user will not look most of the information coming out from cmake.

I made a commit for this and now the reviews have to be made again, I think this is new behavior right? @jgalan @juanangp

juanangp (approved):

Andrii01L (approved):

jgalan:
Yes, I was playing with the settings. If someone pushes new commits, then any approval will be removed, which I believe it makes sense, but if you want, we get back to previous setting? @juanangp @lobis ?

lobis:
> Yes, I was playing with the settings. If someone pushes new commits, then any approval will be removed, which I believe it makes sense, but if you want, we get back to previous setting? @juanangp @lobis ?

I think it makes sense, I agree 👍

PR #187, opened 2022-04-22
Revert "Added preprocessor definition via cmake STRING"

Andrii01L:
Andrii01L Ok: 1 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/revert-185-lobis-regex-lib-compile-def)

Reverts rest-for-physics/framework#185


jgalan:
Why to revert?

Issue #169, opened 2022-03-28
Remove early return on `startup.cpp` when testing is enabled


lobis:
Originally [startup.cpp](https://github.com/rest-for-physics/framework/blob/master/source/framework/core/src/startup.cpp) caused problems with unit testing and to fix this we [added](https://github.com/rest-for-physics/framework/pull/129) an early return if testing was enabled:

```
#ifdef REST_TESTING_ENABLED
        return;
#endif
```

This doesn't seem to affect the running of tests, but makes the framework behave differently when testing is enabled, this means that users should not use the framework with tests compiled to produced results etc. and only use it to run the tests, then compile it again without tests. This is also true for the pipeline where in order to run tests the framework is compiled twice.

There should be a way to modify this file so that tests can be run and also the file keeps working as intended, so that we can avoid recompiling twice.

nkx111:
I am not familiar with ctest. Seems that we can input c++ definitions to the code during the test. Then maybe it is also possible to do something like: 

```
#ifdef REST_TESTING_ENABLED
    REST_COMMIT = REST_TESTING_VAR_COMMIT;
    REST_PATH = REST_TESTING_VAR_PATH;
    REST_USER = REST_TESTING_USER;
#endif
```

where REST_TESTING_XXX are all the imported c++ definitions from the test system.

lobis:
> If returned, the global variables REST_COMMIT/REST_PATH/REST_USER are not set. It will make REST behave differently. I think we should set these variables manually before early return under testing condition.
> 
> ```
> #ifdef REST_TESTING_ENABLED
>     REST_COMMIT = xxxx;
>     REST_PATH = xxxx;
>     REST_USER = xxxx;
>     REST_USER_PATH = xxxx;
>    return;
> #endif
> ```
> 
> What are the anticipated values of these variables?

I am not sure if I follow. The early return is a dirty fix I made in order to avoid tests failing, I think it should be possible to modify this class to remove the early return logic and make it work both during testing and not testing.

nkx111:
> I am not sure if I follow. The early return is a dirty fix I made in order to avoid tests failing, I think it should be possible to modify this class to remove the early return logic and make it work both during testing and not testing.

I was saying that the possible modification is like above. We can do the modification if someone specifies the REST_PATH, etc. values for testing environment.

lobis:
[...]

PR #188, opened 2022-04-26
update logic of REST_StringHelper::isAExpression()

nkx111:
nkx111 Large: 666 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/nkx111-patch-3)

Previously pure number will also be regarded as an expression. In this case long numbers are processed by TFormula and converted into scientific notation, and loses precision.

For example, in the old logic, the parameter:
```
<parameter name="seed" value="123456789"/>
```
will be processed by` REST_StringHelper::EvaluateExpression()`. It will be converted firstly into (double)1.2345789e+08, and then converted into (string)"1.23457e+08", and then back to (int)123457000. This is problematic for seed definition.

So this PR updates the logic of expression identification, inside `REST_StringHelper::isAExpression()`. The simulation results may be different.

restG4 pipeline for branch `nkx111-patch-3` : [](https://gitlab.cern.ch/rest-for-physics/restG4/-/commits/nkx111-patch-3)


jgalan:
It is better to wait for pipeline is green to ask for review. There is a risk the review will be forgotten, because if it is running or red I will change to do something else, and I do not have a new notificiation to remind me I have to review this PR.

jgalan:
Which is the nature of the error at the main framework pipeline? And why the 08.alphas is not present at latest commit at master? @lobis

jgalan:
Should not be simply used `Int_t REST_StringHelper::isANumber(string in)` inside the implementation?

jgalan:
We could perhaps implement `IsAnInteger` that it will be true when the number contains no letters and no `.`.

lobis:
> Which is the nature of the error at the main framework pipeline? @lobis
> 
> In `restG4` why the 08.alphas is not present at latest commit at master?

What do you mean? I see it on `master` https://github.com/rest-for-physics/restG4/tree/master/examples/08.Alphas.

source/framework/tools/src/TRestStringHelper.cxx
```diff
@@ -21,28 +21,56 @@ using namespace std;
 /// \brief Returns 1 only if valid mathematical expression keywords (or numbers)
 /// are found in the string **in**. If not it returns 0.
 ///
+/// By logic, mathematical expressions must have: +-*/e^% in the middle, or % in the end, or math functions in the beginning.
+/// despite those symbols, the string should be purely numeric.
+/// example: 
+/// 1+1 : expression
+/// sin(1.5) : expression
+/// 123456789 : not expression, It is a pure number that can be directly parsed.
 Int_t REST_StringHelper::isAExpression(string in) {
```
lobis, line 30: `REST_StringHelper::isAExpression(const string& in)`

source/framework/tools/src/TRestStringHelper.cxx
```diff
@@ -21,28 +21,56 @@ using namespace std;
...
-    for (int i = 0; i < replace.size(); i++) {
-        temp = Replace(temp, replace[i], "0", 0);
+    bool symbol = false;
+    bool numeric = false;
+    
+    if (in.length() < 2) // minimum expression: 3%
+        return 0;
+
+    vector<string> funcs{"sqrt", "log", "exp", "gaus", "cos", "sin", "tan", "atan", "acos", "asin"};
+    for (int i = 0; i < funcs.size(); i++) {
```
lobis, line 38: for (const auto& item: funcs){
        if (in.find(item) != std::string::npos) {
            symbol = true;
            break;
        }
    }

jgalan:
Ok, it is there at master. Probably I was looking at a branch.

jgalan:
But my question to you was "Which is the nature of the error at the main framework pipeline? How do we debug that?" @lobis

lobis:
> But my question to you was "Which is the nature of the error at the main framework pipeline? How do we debug that?" @lobis

It looks very similar to what I was encountering due to `startup.cpp` when implementing testing. It looks that the error comes from unit testing targets.

The main cause of the problem is still not understood by me and the dirty fix was the ignore some code on `startup.cpp`, this issue https://github.com/rest-for-physics/framework/issues/169 is still opened for whoever wants to fix this. 

My bet for this failure is that maybe some of the changes I made to avoid `startup.cpp` on testing have been reverted perhaps? Or something similar.

jgalan:
But it says `time out` perhaps is trying to connect to a server and gets no response?

lobis:
> But it says `time out` perhaps is trying to connect to a server and gets no response?

Today this has happened a few times, maybe there is an issue with the gitlab or our network

lobis:
I am still not sure why pipeline is failing, but in this PR https://github.com/rest-for-physics/framework/pull/189 something similar seems to be taking place...

Locally everything works but in the pipeline fails.

jgalan (commented):
Should not be simply used Int_t REST_StringHelper::isANumber(string in) inside the implementation? Perhaps a new method `isAnInteger` would be handy?

PR #189, opened 2022-04-27
Event Process Update

lobis:
lobis Large: 650 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/lobis-process-update)

This PR introduces some global changes to the event analysis processes schema.

Many style changes following https://google.github.io/styleguide/cppguide.html such as:

* Made many methods const whenever possible.
* Using `nullptr`.
* Renamed variables for more descriptive names.
* Updated a few method names because of typos such as `GetBytesReaded` -> `GetBytesRead`.
* Added `override` keyword when overriding virtual functions. This also greatly reduces the number of ROOT related warnings!

`TRestEventProcess`:
* Added pure virtual function `GetProcessName`, every process should define this with the process name.
* Added getters/setters for some processes. TODO: perhaps we should have some automated way of doing this? easiest is to declare configurable members public.

Added new example in `pipeline/analysisProcesses` to perform a simple analysis using Python.

Related PRs:

* https://github.com/rest-for-physics/detectorlib/pull/48
* https://github.com/rest-for-physics/geant4lib/pull/44
* https://github.com/rest-for-physics/rawlib/pull/56
* https://github.com/rest-for-physics/connectorslib/pull/11
* https://github.com/rest-for-physics/tracklib/pull/16


source/framework/analysis/inc/TRestMySQLToAnalysisProcess.h
```diff
@@ -89,23 +89,23 @@ class TRestMySQLToAnalysisProcess : public TRestEventProcess {
...
     /// Returns a new instance of this class
     TRestEventProcess* Maker() { return new TRestMySQLToAnalysisProcess; }
 
     /// Returns the name of this process
-    TString GetProcessName() { return (TString) "SQLToAnalysisProces"; }
+    inline TString GetProcessName() const { return (TString) "SQLToAnalysisProces"; }
 
     TRestMySQLToAnalysisProcess();
-    TRestMySQLToAnalysisProcess(char* cfgFileName);
+    TRestMySQLToAnalysisProcess(char* configFilename);
```
juanangp, line 108: Cannot be `TRestMySQLToAnalysisProcess(const char* configFilename);` ?

source/framework/core/inc/TRestEventProcess.h
```diff
@@ -102,75 +102,64 @@ class TRestEventProcess : public TRestMetadata {
...
     //////////////////////////////////////////////////////////////////////////
-    /// \brief Set observable value for analysistree.
+    /// \brief Set observable value for AnalysisTree.
     ///
     /// It will rename the observable to "processName_obsName"
     /// If use dynamic observable, it will try to create new observable
     /// in the AnalysisTree if the observable is not found
     template <class T>
-    void SetObservableValue(std::string name, const T& value) {
+    inline void SetObservableValue(std::string name, const T& value) {
```
juanangp, line 117: In general is not a good practise to have such a long `inline` functions

source/framework/analysis/inc/TRestSummaryProcess.h
```diff
@@ -73,23 +73,23 @@ class TRestSummaryProcess : public TRestEventProcess {
 
    protected:
    public:
-    any GetInputEvent() { return fEvent; }
-    any GetOutputEvent() { return fEvent; }
+    inline any GetInputEvent() const { return fEvent; }
```
juanangp, line 76: I believe is not a good idea to `inline` virtual functions

juanangp:
Looks fine to me although I left some comments.
I think we should also change the inherited functions in libraries and perhaps packages. In that case, can you post the related PR in this PR?

source/framework/core/inc/TRestEventProcess.h
lobis: > Looks fine to me although I left some comments. I think we should also change the inherited functions in libraries and perhaps packages. In that case, can you post the related PR in this PR?

Thanks for the comments, but the PR is still work in progress! I will link everything in this PR when I have finished.

lobis: I agree, but I couldn't change the original implementation, perhaps you know how to do it? remember this is a template function, its not straightforward to define in the cxx.

juanangp: Indeed, I dind't realize that it is a template function. The implementation is relatively easy, however the explicit instantiation is a lot of headache. You can leave as it is.

juanangp:
I just figure out that we had several warnings about `override`. I have pushed a commit to fix it, just some guidelines:
- Use `ClassDeffOverride` instead of `ClassDef`
- Explicit label of override functions with `override`

Check [a00b197](https://github.com/rest-for-physics/framework/pull/189/commits/a00b1975c203973bb8c13a66a3e7175592909e73) for more details.

Please, pay attention for future code updates to try to use the same schema.

lobis:
I have no idea why the pipeline is failing, its failing on the test stage but I have checked locally and every test passes.

It may have to do with the warnings such as:

```
5: Info in <TCling::LoadPCM>: In-memory ROOT PCM candidate /usr/local/root/install/lib/libASImage_rdict.pcm
```

Looks like there are way too many messages like this and maybe this is making the pipeline fail even though the tests passes?

Does anyone @rest-for-physics/core_dev @jgalan @juanangp have any idea how to turn these messages off? Otherwise hopefully the root forum will help us.

juanangp:
> Does anyone @rest-for-physics/core_dev @jgalan @juanangp have any idea how to turn these messages off? Otherwise hopefully the root forum will help us.

Issue seems related with `testRestFramework`
image
All junk appear after this line

lobis:
> > Does anyone @rest-for-physics/core_dev @jgalan @juanangp have any idea how to turn these messages off? Otherwise hopefully the root forum will help us.
> 
> Issue seems related with `testRestFramework` image All junk appear after this line, well in fact appears after linking any executable...

I have no idea how to fix this :(

I tried with `gSystem->RedirectOutput("/dev/null");` with no luck, what puzzles me the most is that these messages appear on the cmake stage AND in the run stage...

juanangp:
I have tried to fix override warnings but the pipeline still failing...

juanangp:
Btw, looking at the error seems that the root pcm files are not properly linked:
`Error in <TCling::LoadPCM>: ROOT PCM /builds/rest-for-physics/framework/build/source/framework/CINT_TRestDataBase_rdict.pcm file does not exist`

The pcm files should be in:
`/builds/rest-for-physics/framework/build/rootdict/`

lobis:
> Btw, looking at the error seems that the root pcm files are not properly linked: `Error in <TCling::LoadPCM>: ROOT PCM /builds/rest-for-physics/framework/build/source/framework/CINT_TRestDataBase_rdict.pcm file does not exist`
> 
> The pcm files should be in: `/builds/rest-for-physics/framework/build/rootdict/`

Thanks! Perhaps you found the issue, now lets try to find out how to fix it...

Also please don't forget to `clang-format` the code before submitting a push 🙏🏻

lobis:
> > > > Does anyone @rest-for-physics/core_dev @jgalan @juanangp have any idea how to turn these messages off? Otherwise hopefully the root forum will help us.
> > > 
> > > 
> > > Issue seems related with `testRestFramework` image All junk appear after this line, well in fact appears after linking any executable...
> > 
> > 
> > Actually I have seen this kind of PCM messages while linking to Garfield. My bet is that ROOT was not able to find the dictionary files. So, at the time, it was probably a problem with the Garfield installation not placing the PCM files somewhere were ROOT would find them, like LD_LIBRARY_PATH.
> 
> When googling I saw you on the forums ^^. I will try to remove Garfield from the pipeline and see what happens, and try to fix it via environment variables... but I thought I had installed everything correctly. I am doing various tests in order to check this. Any idea what I might be doing wrong?
> 
> Its easy to make a mistake because you cannot use an env script such as `thisREST.sh` since the pipeline cannot load this (whichout adding it explicitly), so its possible to make many mistakes.
> 
> > In this case, it is likely an issue with ROOT installation at the docker. Do you reproduce this problem locally? I have never seen those messages, since a long time ago.
> 
> I am using the docker container as my development environment so perhaps I am blind in that regard, perhaps @juanangp has tried in sultan?
> 
> > Was ROOT compiled from source? It looks as it is installed at /usr/local, so it is perhaps installed from a binary?
> 
> `root` is compiled from source.

juanangp:
I have tried in sultan and it compiles properly, however it is not possible to enable the test `-DTEST=ON`.

While using the docker I am able to compile with test enabled.

lobis:
> I have tried in sultan and it compiles properly, however it is not possible to enable the test `-DTEST=ON`.

I don't understand why testing doesn't work on sultan... any ideas are welcome, but as long as it works on the pipeline its ok I think.

juanangp:
> > I have tried in sultan and it compiles properly, however it is not possible to enable the test `-DTEST=ON`.
> 
> I don't understand why testing doesn't work on sultan... any ideas are welcome, but as long as it works on the pipeline its ok I think.

Perhaps the cmake version? It is actually `3.13.4` in sultan
image

lobis:
> > > I have tried in sultan and it compiles properly, however it is not possible to enable the test `-DTEST=ON`.
> > 
> > 
> > I don't understand why testing doesn't work on sultan... any ideas are welcome, but as long as it works on the pipeline its ok I think.
> 
> Perhaps the cmake version? It is actually `3.13.4` in sultan image

I installed a newer version at `/usr/local/bin/cmake`, but I think it still fails

juanangp:
> I installed a newer version at `/usr/local/bin/cmake`, but I think it still fails

With `/usr/local/bin/cmake` I am able to compile with `-DTEST=ON`.
Still, I have several `Info in <TCling::LoadPCM>:` errors while REST is not sourced.
So, I guess the issue is that the `pcm` libraries are not found, because while sourcing REST it is exported to `LD_LIBRARY_PATH`.

juanangp:
I think that I found the issue with the failing pipeline:
image

Looks like `${CI_PROJECT_DIR}/projects/basic-readouts/` is empty, check `ls` output. Actually, `git log` output in this submodule should be like that:
image

Do you understand what is going on? From the build I can check that the submodules are properly built:
image

nkx111:
> > Does anyone @rest-for-physics/core_dev @jgalan @juanangp have any idea how to turn these messages off? Otherwise hopefully the root forum will help us.
> 
> Issue seems related with `testRestFramework` image All junk appear after this line, well in fact appears after linking any executable...

I have fixed this problem by switching the install and test sequence.

https://github.com/rest-for-physics/framework/pull/201#event-6587811932

juanangp:
Looks fine to me, however I can see some warnings on `TRestRawSignalRecoverChannelsProcess` about `override`, @jgalan  this class is deprecated?
Also, we should add the `ClassDefOverride` and the `override` functions to the template macro.

juanangp (approved):

jgalan:
> Looks fine to me, however I can see some warnings on `TRestRawSignalRecoverChannelsProcess` about `override`, @jgalan this class is deprecated?

It is not deprecated. There are recent commits. See for example rest-for-physics/rawlib#57 The class was moved from `connectorslib` to `rawlib`.

jgalan (approved):

juanangp:
> > Looks fine to me, however I can see some warnings on `TRestRawSignalRecoverChannelsProcess` about `override`, @jgalan this class is deprecated?
> 
> It is not deprecated. There are recent commits. See for example [rest-for-physics/rawlib#57](https://github.com/rest-for-physics/rawlib/pull/57) The class was moved from `connectorslib` to `rawlib`.

I see, then perhaps we should fix the warnings before performing the merge, although this can be done later

PR #193, opened 2022-05-03
Jgalan run members

jgalan:
A major update adding a NEW metadata member inside TRestRun and REMOVING an unused data member,

This might cause issues. It would be a good moment to integrate a validation pipeline reading existing ROOT files generated with previous versions of the code.

@rest-for-physics/core_dev

### DIFF
--- a/source/framework/core/inc/TRestRun.h
+++ b/source/framework/core/inc/TRestRun.h
@@ -17,24 +17,44 @@ class TRestEventProcess;
 /// Data provider and manager in REST
 class TRestRun : public TRestMetadata {
    protected:
-    // run info
-    Int_t fRunNumber;  //< first identification number
-    Int_t fParentRunNumber;
-    TString fRunClassName;
-    TString fRunType;  //< Stores bit by bit the type of run. e.g. calibration, background, pedestal,
-                       // simulation, data taking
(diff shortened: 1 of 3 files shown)


jgalan:
Hi, in order to build a validation pipeline, we could have a repository where we upload small files that can be opened with future versions of the code.

If I do not update the class version, then we get the following output

```
Warning in <TStreamerInfo::CompareContent>: The following data member of
the in-memory layout version 5 of class 'TRestRun' is missing from 
the on-file layout version 5:
   TString fRunDescription; //
Warning in <TStreamerInfo::CompareContent>: The following data member of
the in-memory layout version 5 of class 'TRestRun' is missing from 
the on-file layout version 5:
   TString fExperimentName; //
```

I was thinking about capturing that output and check wether it contains the `Warning` keyword. Then the pipeline should fail.

However, I am experiencing problems when capturing the output. If I do:

```
restRoot -q Run00001_NLDBD_Test.root >> output.log
```

It simply does not work, it just captures a partial output, I guess because of redirections of std output.

Any ideas? @nkx @lobis @juanangp @rest-for-physics/core_dev

jgalan:
Any ideas/shortcut on how to capture the output I mentioned above?

PR #198, opened 2022-05-06
Adding pattern mask classes

jgalan:
jgalan Large: 1594 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/jgalan_masks)

This PR adds an abstract metadata class `TRestPatternMask` that encapsulates the implementation of masks (or filters) that define a pattern. The abstract class implements a virtual method `GetRegion` that must be implemented at the inherited class with the logic for each type of mask. This method must return an integer identifying the region where the particle was hit.

`TRestPatternMask` implements a common method to test the mask `DrawMonteCarlo`, generating a canvas with the result.

`TRestPatternMask` implements rotation, pattern shift, and pattern limits (circular mask), those can be applied at each `GetRegion` method by using the method `ApplyCommonMaskTransformation`.

Four types of mask have been implemented at this PR. Stripped, Grid, Rings, Spider.

masks


macros/REST_MakeMetadata.C
```diff
@@ -70,6 +70,8 @@ Int_t REST_MakeMetadata(TString name) {
     headerFile << "    /// REMOVE MEMBER. A dummy member that will be NOT written to the ROOT file." << endl;
     headerFile << "    Double_t fDummyVar = 3.14; //!" << endl;
     headerFile << endl;
+    headerFile << "    void Initialize();" << endl;
```
juanangp, line 73: Don't quite understand this `Initialize`, there is another `public` one below properly tagged as `override`

jgalan: I just moved the method from public to private for some reason I dont recall now.

jgalan: Probably is the conflict resolution that kept both methods. Now I just added the override at the private and removed the public.

lobis (approved):

source/framework/tools/inc/TRestPatternMask.h
```diff
@@ -0,0 +1,91 @@
...
+    /// It defines the mask radius
+    void SetMaskRadius(const Double_t& radius) { fMaskRadius = radius; }
+
+    void PrintMetadata();
+
+    TCanvas* DrawMonteCarlo(Int_t nSamples = 10000);
+
+    ~TRestPatternMask();
+
+    ClassDef(TRestPatternMask, 1);
```
juanangp, line 89: This class inherits from `TRestMetadata` shouldn't be stated as `ClassDefOverrride`. Same applies to the virtual inherited functions that should be label as `override`

source/framework/tools/inc/TRestGridMask.h
```diff
@@ -0,0 +1,62 @@
...
+    /// It returns the modulus used to define a finite set of ids
+    Int_t GetModulus() { return fModulus; }
+
+    void PrintMetadata();
+
+    TRestGridMask();
+    TRestGridMask(const char* cfgFileName, std::string name = "");
+    ~TRestGridMask();
+
+    ClassDef(TRestGridMask, 1);
```
juanangp, line 60: Same here, in this case inherits from `TRestPatternMask` that inherits from `TRestMetadata`  should be also stated as override` override

source/framework/tools/inc/TRestPatternMask.h
jgalan: You probably mean it should be stated as `ClassDefOverride`?

source/framework/tools/inc/TRestGridMask.h
jgalan: Fixed

source/framework/tools/inc/TRestPatternMask.h
jgalan: Fixed

juanangp (approved):

Issue #190, opened 2022-05-02
Adding compression and splitting option to output file


nkx111:
It is interesting to implement compression and auto splitting for process output files. Sometimes the file size becomes too large in large dataset. If it exceeds 100GB I believe there will be problems due to the default TTree size. By either doing file compression and auto splitting we can fix this problem. I have written some code and tested that the auto splitting is working.

1. If file size is small, every thing is normal
2. If it exceeds the limit(set by rml), REST will create extra file series. The splitted files are named as XXX.root.1, XXX.root.2, etc.(with main file named XXX.root). The metadata is still stored in the main file. The splitted files only store analysis tree and event tree object.
3. When opening with TRestRun, if there is extra files series for the input file, It will automatically convert EventTree as TChain. For AnalysisTree, we link additional TChain object, and replace methods like Draw(), Scan(), GetEntries(), etc. to those from TChain.

![image](https://user-images.githubusercontent.com/14999274/166233165-51b8d413-a7d1-4c0d-b059-50300d5d24ed.png)
![image](https://user-images.githubusercontent.com/14999274/166232720-605b9c5b-372c-4d59-8627-ac73528ba7e8.png)

jgalan:
Looks nice! I would then add bytes units to `TRestSystemOfUnits` such as `Mb,kb,Gb,Tb`.

PR #201, opened 2022-05-08
Implement compression and splitting option in TRestProcessRunner

nkx111:
nkx111 Large: 749 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/nkx111-patch-3)

Related to https://github.com/rest-for-physics/framework/issues/190. 

This PR implements compression and splitting option in TRestProcessRunner. The details of these feature can be found in the [issue](https://github.com/rest-for-physics/framework/issues/190). It also introduces other changes:

1. REST now reads correctly long numbers from rml. see [#PR188](https://github.com/rest-for-physics/framework/pull/188). pipelines are updated accordingly.
2. ~~test job in ci is moved after build and install job. This avoids warnings of PCM not find~~(solved in another PR)
3. TRestEventProcess adds a new virtual method `NotifyAnalysisTreeReset()`. If the file is split, this method will be called.
4. TRestThread no longer inherits from TRestMetadata. This saves memory.
5. restRoot initializes event tree and analysis tree as command line object. They are named as ana_tree0, ev_tree0, etc.

restG4 pipeline for branch `nkx111-patch-3` : [](https://gitlab.cern.ch/rest-for-physics/restG4/-/commits/nkx111-patch-3)


pipeline/selection/Validate.C
```diff
@@ -7,14 +7,16 @@ Int_t Validate(TString fname) {
...
+
+    if (tr->GetEntries() != 1) {
         cout << "Number of entries is not the same!" << endl;
-        cout << "Expected: 3. Obtained: " << tr->GetEntries() << endl;
+        cout << "Expected: 1. Obtained: " << tr->GetEntries() << endl;
         return 1;
     }
 
     // Check IDs.
-    std::vector<Int_t> ids = {0, 3, 6};
```
jgalan, line 17: Why the events selected is not the same now? @DavidDiezIb

source/bin/restRoot.cxx
```diff
@@ -103,6 +103,36 @@ int main(int argc, char* argv[]) {
                 runTmp->GetEntry(0);
             }
 
+            // command line AnalysisTree object
```
jgalan, line 106: This addon was not mentioned in the PR description

pipeline/selection/g4OnSelection.rml
```diff
@@ -26,7 +26,7 @@
  
     <addProcess type="TRestEventSelectionProcess" name="evSelection" fileWithIDs="Run00001_NLDBD_Test_g4Analysis.root" conditions="g4Ana_totalEdep<2400 AND g4Ana_energyPrimary>1000" value="ON"  verboseLevel="info"/>
     
-    <addProcess type="TRestEventSelectionProcess" name="evSelection" fileWithIDs="IDs.txt" value="ON"  verboseLevel="info"/>
+    <!--<addProcess type="TRestEventSelectionProcess" name="evSelection" fileWithIDs="IDs.txt" value="ON"  verboseLevel="info"/>-->
```
jgalan, line 29: why this was commented out? It was not said at the PR description.

source/bin/restRoot.cxx
nkx111: OK, I updated the description.

pipeline/selection/g4OnSelection.rml
nkx111: It is mentioned that some pipelines are updated.

In principle I think adding two TRestEventSelectionProcess is not a good pipeline script. We shall run two indepent processes to test "condition selection" and "ids file selection" functionality.

jgalan: Probably then we need now to recover the `IDs.txt` event selection pipeline? @DavidDiezIb

DavidDiezIb:
I splited the EventSelectionProcess pipeline in two parts, with different rml and validation files.

jgalan (approved):

.gitlab-ci.yml
```diff
@@ -112,6 +95,24 @@ Build and Install:
       - ${CI_PROJECT_DIR}/install
     expire_in: 1 day
 
+Build and Test:
+  stage: test
+  script:
+    - source ${CI_PROJECT_DIR}/install/thisREST.sh
```
juanangp, line 101: There is already a solution for this issue on #202.
On this other hand, I don't think that sourcing a compiled version of `rest` without the `testing` enabled is the way to go.

source/bin/restRoot.cxx
```diff
@@ -103,6 +103,36 @@ int main(int argc, char* argv[]) {
...
+                string evcmd = Form("TRestAnalysisTree* ana_tree%i = (TRestAnalysisTree*)%s;", Nfile,
+                                    ToString(runTmp->GetAnalysisTree()).c_str());
+                if (debug) printf("%s\n", evcmd.c_str());
+                gROOT->ProcessLine(evcmd.c_str());
+                // runTmp->GetEntry(0);
+                //}
+            }
+
+            // command line EventTree object
+            if (runTmp->GetEventTree() != NULL) {
```
juanangp, line 127:
```suggestion
            if (runTmp->GetEventTree() != nullptr) {
```

source/framework/core/inc/TRestAnalysisTree.h
```diff
@@ -128,6 +136,9 @@ class TRestAnalysisTree : public TTree {
             std::cout << "Error! TRestAnalysisTree::GetObservableValue(): index outside limits!" << endl;
             return T();
         }
+        if (fChain != NULL) {
```
juanangp, line 139:
```suggestion
        if (fChain != nullptr) {
```

source/framework/core/src/TRestAnalysisTree.cxx
```diff
@@ -982,6 +1004,88 @@ Int_t TRestAnalysisTree::WriteAsTTree(const char* name, Int_t option, Int_t bufs
     return result;
 }
 
+/// <summary>
+/// Add a series output file like TChain.
+/// </summary>
+/// <param name="file"> The input file that contains another AnalysisTree with same run id </param>
+/// <returns></returns>
+Bool_t TRestAnalysisTree::AddChainFile(string _file) {
+    TFile* file = new TFile(_file.c_str(), "read");
```
juanangp, line 1013: There is not reason to call `new` in this case.
```suggestion
    TFile file (_file.c_str(), "read");
```
or
```suggestion
   auto file = std::unique_ptr<TFile> {TFile::Open(_file.c_str(), "read")};
```

source/framework/core/src/TRestAnalysisTree.cxx
```diff
@@ -982,6 +1004,88 @@ Int_t TRestAnalysisTree::WriteAsTTree(const char* name, Int_t option, Int_t bufs
...
+        warning << "TRestAnalysisTree::AddChainFile(): failed to open file " << _file << endl;
+        return false;
+    }
+
+    TRestAnalysisTree* tree = (TRestAnalysisTree*)file->Get("AnalysisTree");
+    if (tree != nullptr && tree->GetEntry(0)) {
+        if (tree->GetEntry(0) > 0) {
+            if (tree->GetRunOrigin() == this->GetRunOrigin()) {
+                // this is a valid tree
+                delete tree;
```
juanangp, line 1024: I don't think is a good practise to `delete` an object which has not been created with `new`, you can fall in an unexpected behaviour.

source/framework/core/src/TRestProcessRunner.cxx
```diff
@@ -211,14 +223,18 @@ Int_t TRestProcessRunner::ReadConfig(string keydeclare, TiXmlElement* e) {
...
-                        fThreadNumber--;
+                    if (fThreadNumber > 1) {
+                        info << "multi-threading is disabled due to process \"" << p->GetName() << "\""
+                             << endl;
+                        info << "This process is in debug mode or is single thread only" << endl;
+                        for (i = fThreadNumber; i > 1; i--) {
+                            // delete (*fThreads.end());
+                            fThreads.erase(fThreads.end() - 1);
+                            fThreadNumber--;
+                        }
```
juanangp, line 237: Don't quite understand this is this bucle for? Note that you have to `delete` the pointer content, if not you have a potential memory leak.
```suggestion
                        for (auto it = fThread.begin(); it != fThread.end(); ) {
                           if( it !=fThread.begin() ) {
                             delete * it;
                             it = fThread.erase(it);
                             fThreadNumber--;
                           } else {
                             ++it;
                           }
                        }
```

source/framework/core/src/TRestProcessRunner.cxx
```diff
@@ -824,6 +844,74 @@ void TRestProcessRunner::FillThreadEventFunc(TRestThread* t) {
...
+                     << " bytes), switching to new file with index " << fNFilesSplit << endl;
+
+                // wait 0.1s for the process to finish
+                usleep(100000);
+                for (int i = 0; i < fThreadNumber; i++) {
+                    for (int j = 0; j < fProcessNumber; j++) {
+                        auto proc = fThreads[i]->GetProcess(j);
+                        proc->NotifyAnalysisTreeReset();
+                    }
+                }
```
juanangp, line 863:
```suggestion
                for (auto th : fThreads) {
                    for (int j = 0; j < fProcessNumber; j++) {
                        auto proc = th->GetProcess(j);
                        proc->NotifyAnalysisTreeReset();
                    }
                }
```

source/framework/core/src/TRestProcessRunner.cxx
```diff
@@ -824,6 +844,74 @@ void TRestProcessRunner::FillThreadEventFunc(TRestThread* t) {
...
+                    fEventTree->Reset();
+                }
+
+                // write some information to the first(main) data file
+                fRunInfo->SetNFilesSplit(fNFilesSplit);
+                if (fOutputDataFile->GetName() != fOutputDataFileName) {
+                    TFile* Mainfile = new TFile(fOutputDataFileName, "update");
+                    WriteMetadata();
+                    Mainfile->Close();
+                    delete Mainfile;
```
juanangp, line 879: Still no need to use `new` and `delete` inside the if.
```suggestion
                    auto Mainfile = std::unique_ptr<TFile> {TFile::Open(fOutputDataFileName, "update")};
                    WriteMetadata();
                    Mainfile->Close();
```

source/framework/core/src/TRestProcessRunner.cxx
```diff
@@ -824,6 +844,74 @@ void TRestProcessRunner::FillThreadEventFunc(TRestThread* t) {
...
+                } else {
+                    WriteMetadata();
+                }
+
+                TFile* newfile = new TFile(fOutputDataFileName + "." + ToString(fNFilesSplit), "recreate");
+
+                TBranch* branch = 0;
```
juanangp, line 886:
```suggestion
                TBranch* branch = nullptr;
```

.gitlab-ci.yml
nkx111: > There is already a solution for this issue on https://github.com/rest-for-physics/framework/pull/202.

I see. It is fixed in 172ba772. No need to change pipeline now.

> On this other hand, I don't think that sourcing a compiled version of rest without the testing enabled is the way to go.

I don't understand. Does `tesing` make any difference to the compilation output? In principle it is compiling additional executables based on same library. So the PCM files can be reused.

source/framework/core/inc/TRestAnalysisTree.h
```diff
@@ -266,6 +281,93 @@ class TRestAnalysisTree : public TTree {
...
+
+    Long64_t GetEntries(const char* sel);
+
+    void Browse(TBrowser* b) { fChain ? fChain->Browse(b) : TTree::Browse(b); }
+    Long64_t Draw(const char* varexp, const TCut& selection, Option_t* option = "",
+                  Long64_t nentries = kMaxEntries, Long64_t firstentry = 0) {
+        return fChain ? fChain->Draw(varexp, selection, option, nentries, firstentry)
+                      : TTree::Draw(varexp, selection, option, nentries, firstentry);
+    }
+    Long64_t Draw(const char* varexp, const char* selection, Option_t* option = "",
```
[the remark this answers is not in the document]
nkx111: Yes. It is used to read splitted data files. In case multiple files are created to store analysis tree data, we make TRestAnalysisTree running internally with TChain. All the functions called to TRestAnalysisTree is calling back the embed TChain object.

source/framework/core/src/TRestAnalysisTree.cxx
nkx111: But this is what returned by ROOT. We can only trust that TFile::Get() returns the pointer which is deleteable.

source/framework/core/src/TRestProcessRunner.cxx
nkx111: I don't know why but it always causes problems when I delete the TRestThread object. This code block is for a rare case that, when adding processes, one process turns out to be must running under single thread. So we remove other threads. I guess this memory leak is acceptable and so I directly comment the delete operation.

source/framework/core/src/TRestAnalysisTree.cxx
juanangp: For sure you can delete the pointer, however `root` has his own memory management, which means that it can try to access de `tree` that you just delete, which can end up in a seg fault.

source/framework/core/src/TRestProcessRunner.cxx
juanangp: But it was deleted in the previous implementation...

source/framework/core/src/TRestAnalysisTree.cxx
nkx111: I don't understand. In which case will ROOT access such object? Or there is any "safe" method provided by ROOT to delete it?

source/framework/core/src/TRestProcessRunner.cxx
nkx111: In the previous implementation this delete operation sometimes causes seg. fault which I don't know why.

source/framework/core/src/TRestAnalysisTree.cxx
juanangp: The `tree` in fact belongs to the `TFile`, when the `TFile` is destroyed it migth try to `delete` the tree, so you can fall in a double free with an unexpected behaviour.
https://stackoverflow.com/questions/9169774/what-happens-in-a-double-delete

source/framework/core/src/TRestProcessRunner.cxx
juanangp: Now that I check more thoroughly, looks like you have a double bucle for `fThreads`, which can be the origin of the issue.
```
for (int i = 0; i < fThreadNumber; i++) {
....
               for (i = fThreadNumber; i > 1; i--)
```
I would suggest to use std iterators instead:
```
for (auto it = fThread.begin(); it != fThread.end(); ) {
```

source/framework/core/src/TRestAnalysisTree.cxx
nkx111: I did a test and there is no such double free fault if I delete the tree manually. There is also no memory leak if I don't delete the tree. So I guess ROOT has some mechanism to prevent object from file being deleted. Either delete or not is OK.

nkx111: But I still need to delete the tree. Otherwise the initializaion of the TChain will be problematic.

nkx111: @juanangp Can you approve this PR? I believe those problems are trivial

juanangp (approved):

PR #202, opened 2022-05-09
Minor updates for TRestGDMLParser

lobis:
lobis Medium: 147 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/lobis-activeVolume-43)

* Related to https://github.com/rest-for-physics/restG4/issues/43
* Replaces Nevents by nEvents in RML files (Nevents still works) for consistency
* Replaces gdml_file by gdmlFile in RML files (gdml_file still works) for consistency

TRestGDMLParser:
* Renamed members for consistency (starting with `f`...)
* Made some methods parameters by const ref.
* When USER is not defined, will save temporary gdml file to `/tmp` directory instead of failing
* Includes <filesystem> header to create directory recursively if not exists
* Other minor changes

### DIFF
--- a/cmake/thisREST.cmake
+++ b/cmake/thisREST.cmake
@@ -15,7 +15,7 @@ string(REGEX REPLACE "\n$" "" GEANT4_PATH "${GEANT4_PATH}")
 set(thisGeant4 "${GEANT4_PATH}/bin/geant4.sh")
 
 if (${REST_G4} MATCHES "ON")
-    set(loadG4 "\# if geant4.sh script is found we load the same Geant4 version as used in compilation\n if [[ -f \"${thisGeant4}\" && ${thisGeant4} != /usr/* ]]; then
+    set(loadG4 "\# if geant4.sh script is found we load the same Geant4 version as used in compilation\n if [[ -f \\\"${thisGeant4}\\\" && ${thisGeant4} != /usr/* ]]; then
 		source ${thisGeant4}\n fi\n")
 else ()
     set(loadG4 "")
--- a/source/framework/core/inc/TRestGDMLParser.h
+++ b/source/framework/core/inc/TRestGDMLParser.h
@@ -39,6 +39,9 @@ class TRestGDMLParser : public TRestMetadata {
     }
 
     TGeoManager* CreateGeoManager();
+    [[deprecated("Use CreateGeoManager() instead.")]] inline TGeoManager* CreateGeoM() {
+        return CreateGeoManager();
+    }
 
     void InitFromConfigFile() override {}
 
--- a/source/framework/core/inc/TRestGDMLParser.h
+++ b/source/framework/core/inc/TRestGDMLParser.h
@@ -13,30 +13,30 @@
 
 class TRestGDMLParser : public TRestMetadata {
    private:
-    TGeoManager* fGeo;
+    TGeoManager* fGeoManager;
 
    public:
     TRestGDMLParser() {}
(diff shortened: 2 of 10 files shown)


cmake/thisREST.cmake
```diff
@@ -15,7 +15,7 @@ string(REGEX REPLACE "\n$" "" GEANT4_PATH "${GEANT4_PATH}")
 set(thisGeant4 "${GEANT4_PATH}/bin/geant4.sh")
 
 if (${REST_G4} MATCHES "ON")
-    set(loadG4 "\# if geant4.sh script is found we load the same Geant4 version as used in compilation\n if [[ -f \"${thisGeant4}\" && ${thisGeant4} != /usr/* ]]; then
+    set(loadG4 "\# if geant4.sh script is found we load the same Geant4 version as used in compilation\n if [[ -f \\\"${thisGeant4}\\\" && ${thisGeant4} != /usr/* ]]; then
```
jgalan, line 18: is this a bug fix?

source/framework/core/inc/TRestGDMLParser.h
```diff
@@ -7,49 +7,51 @@
...
 
     void PrintContent();
 
     void ReplaceEntity();
 
-    void ReplaceAttributeWithKeyWord(std::string keyword);
+    void ReplaceAttributeWithKeyWord(const std::string& keyword);
 
-    ClassDef(TRestGDMLParser, 1);
+    ClassDefOverride(TRestGDMLParser, 2);
```
jgalan, line 54: For curiosity, when should we use `DefOverride` and why?

source/framework/core/src/TRestMetadata.cxx
```diff
@@ -515,8 +515,8 @@ TRestMetadata::TRestMetadata(const char* cfgFileName) : endl(this) {
 /// \brief TRestMetadata default destructor
 ///
 TRestMetadata::~TRestMetadata() {
-    if (fElementGlobal) delete fElementGlobal;
-    if (fElement) delete fElement;
+    delete fElementGlobal;
```
jgalan, line 518: why not this check? It is not bad to check if the pointer is `nullptr` then just do nothing in that case.

jgalan (approved):

juanangp (approved):

cmake/thisREST.cmake
lobis: Not really sure, it was introduced by @juanangp

source/framework/core/src/TRestMetadata.cxx
lobis: > why not this check? It is not bad to check if the pointer is `nullptr` then just do nothing in that case.

Deleting a null pointer is perfectly safe, so there is not need to check this.

https://stackoverflow.com/questions/615355/is-there-any-reason-to-check-for-a-null-pointer-before-deleting

cmake/thisREST.cmake
juanangp: Yes, this is a bug fix, it was giving an error after performing `cmake`

source/framework/core/src/TRestMetadata.cxx
jgalan: The main problem is that delete will not remove the actual pointer address. This is something I discussed with @juanangp already in 2015. So, it is necessary to set it to null to make sure you wont access it again or you don't believe it is existing. Probably at the constructor it does nothing.

PR #203, opened 2022-05-10
Improving clang formatting

juanangp:
juanangp Ok: 87 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/clang-script)

Script `reformat-clang` has been improved:
- Added `readlink` to make the script independent from the path where is launched
- Adding support for single files including file extensions `*.h`, `*.cxx`, `*.cc` and `*.C`
- Now the script check if the file or directory exists
- Removing un-necessary commands

**Added  `githook` in the `cmake`, so the `clang-format` is performed when `git commit` is issued**
- `githook` is created when `cmake` is issued and if there is no `pre-commit` present.
- Check if `clang-format` is present in the system, otherwise is skipped.
- Can be used as template to add different `githook` features.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -126,6 +126,9 @@ string(TIMESTAMP date "%Y-%m-%d %H:%M")
 # Enable testing (-DTEST=ON flag, it is OFF by default)
 include(Testing)
 
+# Includes githooks, now using clang-format
+include(GitHooks)
+
 # Start compile
 include(MacroRootDict)
 add_subdirectory(source)
--- a/cmake/GitHooks.cmake
+++ b/cmake/GitHooks.cmake
@@ -9,7 +9,7 @@ endif ()
 # Installing clang-format pre-commit hook if prerequisites are met and it doesn't exist yet.
 
 if (NOT EXISTS ${PROJECT_SOURCE_DIR}/.git)
-    message(WARNING ".git directory not found at the root of the repository (${PROJECT_SOURCE_DIR}), skipping formatting in pre-commit hook")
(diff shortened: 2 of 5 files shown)


lobis (approved):

jgalan (approved):
As I commented to @lobis it would be interesting that our `Cmake` system forces that `reformat-clang.sh` is executed at the beginning of each `make` execution. So that any user that launches code compilation keeps a good formatting.

lobis:
> As I commented to @lobis it would be interesting that our `Cmake` system forces that `reformat-clang.sh` is executed at the beginning of each `make` execution. So that any user that launches code compilation keeps a good formatting.

It could be nice but maybe it would give problems, also it takes a fair amount of time to execute the clang-format. It would be annoying for users such as me that have this done automatically on file save. I think it may be better a step on the pipeline that gives error if the source files have not been clang-formatted? (calculating the diff etc.).

What do you think @jgalan @juanangp ?

juanangp:
> What do you think @jgalan @juanangp ?
Perhaps would be better to use a pre-commit hook https://stackoverflow.com/questions/55965712/how-do-i-add-clang-formatting-to-pre-commit-hook
But I am not sure how to implement it in the repo...

lobis:
> > What do you think @jgalan @juanangp ?
> > Perhaps would be better to use a pre-commit hook https://stackoverflow.com/questions/55965712/how-do-i-add-clang-formatting-to-pre-commit-hook
> > But I am not sure how to implement it in the repo...

It is not possible to implement it in the repo, at most you can have a file in the repo and direct users to copy this file into their `.git` folder.

juanangp:
> It is not possible to implement it in the repo, at most you can have a file in the repo and direct users to copy this file into their `.git` folder.
Can this be done at the build or in the cmake? Or this is an overkill? Also note that you need `clang-format` installed in your OS

Issue #205, opened 2022-05-11
Format rml files in a similar way as clang-format


juanangp:
Formatting of `rml` files is different across the framework, would be nicer if all the files has the same format, similar to `clang-format`. Two different tasks should be addressed in this issue:
- [x]  Add `rml` formating in[ `reformat-clang.sh`](https://github.com/rest-for-physics/framework/blob/master/scripts/reformat-clang.sh) script.
- [x] Include `rml` formatting in `githooks`

Check https://github.com/rest-for-physics/framework/pull/203 for more details about `githook` and recent changes in `reformat-clang.sh`.

PR #208, opened 2022-05-12
Format rml files in a similar way as clang format

lobis:
lobis Ok: 50 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/205-format-rml-files-in-a-similar-way-as-clang-format)

Fixes https://github.com/rest-for-physics/framework/issues/205


juanangp:
Shall we drop the support for directories in this script? I think, shouldn't bee necessary after implementing `hooks` #209

juanangp:
Shall we drop the support for directories in this script? I think, shouldn't bee necessary after implementing `hooks` #209

lobis:
> Shall we drop the support for directories in this script? I think, shouldn't bee necessary after implementing `hooks` #209

I don't think so, people (like me) would still want to use it for formatting the whole project tree, and I cannot use this cmake hook due to remote build :(

juanangp:
> I don't think so, people (like me) would still want to use it for formatting the whole project tree, and I cannot use this cmake hook due to remote build :(

Ok, fair enough

juanangp (approved):

jgalan (approved):

Issue #217, opened 2022-05-19
CMake preprocessor definitions for libraries not working correctly 


lobis:
When a library is set to be compiled, a preprocessor definition is defined (see https://github.com/rest-for-physics/framework/blob/master/cmake/MacroRootDict.cmake#L523) to tell other libraries that the corresponding library exists.

For example if the detector library is found, `REST_DetectorLib` will be defined as a predecesor directive and this can be used in another library to add some code that would otherwise fail.

However the current implementation is not working correctly due to some bug, as pointed out by @KonradAltenmueller.

PR #218, opened 2022-05-19
Fix library preprocessor directives failing

lobis:
lobis Ok: 24 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/217-cmake-preprocessor-definitions-for-libraries-not-working-correctly)

Fixes https://github.com/rest-for-physics/framework/issues/217.

The preprocessor directives were failing because the libraries were being compiled before some of them were defined. The solution was to define these directives before compilation for all libraries.


KonradAltenmueller:
The pipeline failed... In my installation I could build it without problem.

KonradAltenmueller:
Anyways, I still get the same errors as before when I run a rml with `TRestRawSignalRecoverChannelsProcess`:
```

-- Error : TRestRawSignalRecoverChannelsProcess will not be active.
-- Error : REST was not compiled with detectorlib
-- Error : Please, remove this process or compile REST with detector library
```

jgalan:
Aside from the fact that we need to fix the problem, I suggest to exclude the process in case DetectorLib is not included in the compilation.

I.e. adding `set ( excludes ${excludes} ....` inside the CMakeLists.txt` as follows:
```
set(LibraryVersion "1.6")
add_definitions(-DLIBRARY_VERSION="${LibraryVersion}")

if (NOT ${RESTLIB_DETECTOR} MATCHES "ON")
    set(excludes ${excludes} TRestRawSignalRecoverChannelsProcess)
endif ()

COMPILELIB("")

file(GLOB_RECURSE MAC "${CMAKE_CURRENT_SOURCE_DIR}/macros/*")
INSTALL(FILES ${MAC} DESTINATION ./macros/raw)

ADD_LIBRARY_TEST()
```

lobis:
> @lobis did you try to change `add_compile_definitions` by `add_definitions`?
> 
> We use that here for example
> ```
> if (${SQL} MATCHES "ON")
>     list(APPEND rest_features "SQL")
>     add_definitions(-DUSE_SQL)
>     include(FindSQL)
> ```
> 

Yes, this is the old way of doing this, nowadays it's recommended to use `add_compile_definitions`. The effect is the same. The problem is that the definitions were being inserted after compilation in some cases.

jgalan:
Hi @lobis, I placed myself in your branch and I do not see the library being listed as usual at the Cmake execution output

```

REST will be installed in: /usr/local
===========================================================
System is: Linux
Packages: 
Libraries: RestFramework
Features: Eve
===========================================================
Check main project README.md to see all available compilation options, packages and libraries

-- Configuring done
-- Generating done
-- Build files have been written to: /home/jgalan/rest-framework/build
jgalan@sultan:~/rest-framework/build$ cmake -DRESTLIB_RAW=ON ..
```

lobis:
> Hi @lobis, I placed myself in your branch and I do not see the library being listed as usual at the Cmake execution output
> 
> ```
> 
> REST will be installed in: /usr/local
> ===========================================================
> System is: Linux
> Packages: 
> Libraries: RestFramework
> Features: Eve
> ===========================================================
> Check main project README.md to see all available compilation options, packages and libraries
> 
> -- Configuring done
> -- Generating done
> -- Build files have been written to: /home/jgalan/rest-framework/build
> jgalan@sultan:~/rest-framework/build$ cmake -DRESTLIB_RAW=ON ..
> ```

This change will require more work than expected, perhaps we should just make a patch for the raw + detector case?

jgalan:
I think I found a fix for the directives, the problem is that when it is finally fixed, the code from `TRestDetectorXYZ` is exposed and it will fail compilation because we require to add the linking headers from detectorlib to rawlib.

```
In file included from /home/jgalan/rest-framework/source/libraries/raw/src/TRestRawSignalChannelActivityProcess.cxx:99:
/home/jgalan/rest-framework/source/libraries/raw/inc/TRestRawSignalChannelActivityProcess.h:29:10: fatal error: TRestDetectorReadout.h: No such file or directory
 #include <TRestDetectorReadout.h>
```

So, probably the best choice is to move those classes back to conntectorslib, i.e. `TRestRawSignalChannelActivityProcess` and `TRestRawSignalRecoverChannelsProcess`?

/polls "Add dependency to the library as needed" "Move those classes that require dependencies to connectorslib"

@rest-for-physics/core_dev @juanangp @lobis @nkx111

juanangp:
> So, probably the best choice is to move those classes back to conntectorslib, i.e. `TRestRawSignalChannelActivityProcess` and `TRestRawSignalRecoverChannelsProcess`?

I agree with this.

jgalan:
I added a fix at #222. I tested and it seems to be working.

Then, I just added detectorlib to rawlib compilation as a dependency. See rest-for-physics/rawlib#62.

You could reject rest-for-physics/rawlib#62 and move the required classes back to connectorslib.

PR #221, opened 2022-05-23
Adding new `Angle` unit

lobis:
This PR adds a new unit to measure angles. Previously they were defined only in radians.

Radians is still the default unit, but now `deg` or `degree(s)` can also be used. I also added `rad` and `radian(s)` as units in case anyone wants to be more explicit with the unit.

I will also updated some examples on `restG4` to make use of this unit so that they are easier to understand.

### DIFF
--- a/source/framework/tools/inc/TRestSystemOfUnits.h
+++ b/source/framework/tools/inc/TRestSystemOfUnits.h
@@ -33,7 +33,7 @@
 namespace REST_Units {
 
 // We use more common physics units instead of SI unit
-enum Physical_Unit { Energy, Time, Distance, Mass, Electric, Magnetic, Pressure, Angle, NOT_A_UNIT = -1 };
+enum Physical_Unit { Energy, Time, Length, Mass, Electric, Magnetic, Pressure, Angle, NOT_A_UNIT = -1 };
 
 class TRestSystemOfUnits {
    private:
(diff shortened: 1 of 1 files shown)


source/framework/tools/inc/TRestSystemOfUnits.h
```diff
@@ -23,16 +24,16 @@
 #include <string>
 
 #ifdef REST_UnitsAdd_Caller
-#define AddUnit(name, type, scale) const double name = _AddUnit(#name, type, scale)
+#define AddUnit(name, type, scale) double name = _AddUnit(#name, type, scale)
```
juanangp, line 27: I know this is not related with this PR, but do we really need a macro for that? I think we just need a lookup table

```
const std::map REST_Units_Map<std::string, std::pair<Physical_Unit, double> >{
  {"meV", std::make_pair (REST_Units::Energy, 1e6) },
  {"keV", std::make_pair (REST_Units::Energy, 1e3) },
  ....
}
```
Then to check for a particular unit:
```
if(REST_Units_Map.find("myUnit") == REST_Units_Map.end() ) return error;//Unit not found
Physical_Unit unit =  REST_Units_Map["myUnit"].first;
double conversion = REST_Units_Map["myUnit"].second;
```

jgalan: I dont think it is a systematic practice in our code

lobis: > I dont think it is a systematic practice in our code

I agree, in this case I think it's very well contained.

lobis: why did you request I remove the `const` qualifier @juanangp ?

juanangp:
Most of the pipelines are failing, do you understand why?

lobis:
> Most of the pipelines are failing, do you understand why?

No, but hopefully it gets fixed by merging master intro this branch, as it looks like it has to do with submodules.

source/framework/tools/inc/TRestSystemOfUnits.h
juanangp: I didn't request to remove any `const` under this PR, did I?

lobis: > I didn't request to remove any `const` under this PR, did I?

nevermind, I got it wrong, sorry.

juanangp (approved):

glorialuzon (approved):

PR #225, opened 2022-05-24
Renaming endl to RESTendl and other loggers functions

juanangp:
juanangp Large: 1276 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/RESTendl)

This PR try to address the misleading use of custom `endl` and renames the different custom logger functions:
- Renamed custom `endl` to `RESTendl`. Note that we have 2 different ´RESTendl´. `endl_t RESTendl` inside `TRestMetadata` and `static RESTendl` inside `TRestStringOuput`. However, the implementation has not been changed.
- Renamed other logger functions
  - `info` --> `RESTInfo`  
  - `debug` --> `RESTDebug`
  - `essential` --> `RESTEssential`
  - `warning` --> `RESTWarning`
  - `fout` --> `RESTcout`
  - `ferr` --> `RESTError`
- Added `REST_Verbose_Level` enum inside `TRestStringOuput` class, now verbosity is written as `TRestStringOutput::REST_Verbose_Level::REST_XXXX`.
-  Added `REST_Display_Orientation` enum inside  `TRestStringOuput.h`
- Removal of extern functions from `startup.cpp` now the custom loggers are defined as `static` functions inside `TRestStringOuput.h`
- Removal of `define COLOR_XXX` inside `TRestStringOuput.h`, using `constexpr const char* const COLOR_` instead.

Related PR:
- https://github.com/rest-for-physics/detectorlib/pull/49
- https://github.com/rest-for-physics/geant4lib/pull/49
- https://github.com/rest-for-physics/rawlib/pull/63
- https://github.com/rest-for-physics/tracklib/pull/17
- https://github.com/rest-for-physics/connectorslib/pull/13
- https://github.com/rest-for-physics/axionlib/pull/24
- https://github.com/rest-for-physics/restG4/pull/47

### DIFF
--- a/macros/REST_CheckRunFileList.C
+++ b/macros/REST_CheckRunFileList.C
@@ -20,7 +20,7 @@ Int_t REST_CheckRunFileList(TString namePattern, Int_t N = 100000) {
 
     vector<TString> filesNotWellClosed;
 
-    TRestStringOutput RESTcout;
+    TRestStringOutput RESTLog;
 
     string a = TRestTools::Execute((string)("ls -d -1 " + namePattern));
     vector<string> b = Split(a, "\n");
@@ -29,14 +29,14 @@ Int_t REST_CheckRunFileList(TString namePattern, Int_t N = 100000) {
     int cont = 0;
     for (int i = 0; i < b.size(); i++) {
         string filename = b[i];
-        RESTcout << filename << RESTendl;
+        RESTLog << filename << RESTendl;
         cont++;
         TRestRun* run = new TRestRun();
 
         TFile* f = new TFile(filename.c_str());
 
         if (!TRestTools::fileExists(filename)) {
-            RESTcout << "WARNING. Input file does not exist" << RESTendl;
+            RESTLog << "WARNING. Input file does not exist" << RESTendl;
             exit(1);
         }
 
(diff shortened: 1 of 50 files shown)


jgalan:
what it is doing `fout`? 

Perhaps `ferr` could just be `RESTError`?

juanangp:
> what it is doing `fout`?

I think is standard output with silent verbosity.

> Perhaps `ferr` could just be `RESTError`?

Yes, perhaps we should have a quorum about this? Currently we have:
- `RESTFout` verboseLevel >=  silent --> `RESTCout` ?
- `RESTFerr` verboseLevel >=  silent --> `RESTError` ?
- `RESTWarning` verboseLevel >=  warning 
- `RESTEssential` verboseLevel >=   essential == warning --> `RESTLog` ?
- `RESTInfo` verboseLevel >=  info
- `RESTSuccess`  verboseLevel >=  info
- `RESTDebug`  verboseLevel >=  debug
- `RESTExtreme`  verboseLevel >=  extreme

jgalan:
Ok, I updated `RESTFerr` because it was kind of obvious when comparing with `RESTWarning`.

I agree with `RESTCout` but not sure about `RESTLog` because this is connected to a `verboseLevel="essential"` parameter.

juanangp:
> Ok, I updated `RESTFerr` because it was kind of obvious when comparing with `RESTWarning`.
> 
> I agree with `RESTCout` but not sure about `RESTLog` because this is connected to a `verboseLevel="essential"` parameter and corresponding output level.

OK, I can change that btw, do you prefer `RESTCout` or `RESTcout` without capital letter.

lobis:
I think maybe it would be better to use a namespace for the logger methods i.e. `logger::info` instead of `RESTInfo`. What do you think?

Also I think the logger needs a bit of a rework (on a future PR), some of the things I would modify:

Stop using `cout` like interface for logging messages, use something like https://github.com/fmtlib/fmt instead (probably its alerady packed into root?).

This would allows us, among many things, to define custom formatting for things such as `TVector3` or even REST classes so that we can "just print them", reducing significantly the code.

Use https://github.com/gabime/spdlog as the logging library and build a logger on top of it, instead of implementing it from scratch, which has many nice features including https://github.com/gabime/spdlog#backtrace-support

jgalan (approved):

lobis (approved):

PR #226, opened 2022-05-24
CMakeLists.txt adding std++fs linking

jgalan:
jgalan Ok: 1 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/jgalan_build_fix)

I found out that I need to add this fix in order to be able to compile at sultan2, strange ... but

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -80,7 +80,7 @@ find_package(ROOT REQUIRED COMPONENTS ${ROOT_REQUIRED_LIBRARIES})
 
 message(STATUS "ROOT LIBRARIES: ${ROOT_LIBRARIES}")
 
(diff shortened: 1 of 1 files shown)


lobis:
> jgalan Ok: 1 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/jgalan_build_fix)
> 
> I found out that I need to add this fix in order to be able to compile at sultan2, strange ... but

This is because we are using `gcc` version `8`, which I think its fairly old? (but still recieving updates?), I am not sure how it works, but for example in my dev container using `Ubuntu 22.04` the default is version `11`.

lobis (approved):

jgalan:
It is always good to keep compatibility with as older compiler as possible since this will allow REST to be installed at any system.

PR #229, opened 2022-05-26
Simplified some of TRestTools methods using `<filesystem>` and avoid usage of env variables

lobis:
lobis Medium: 104 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/lobis-filesystem)

Before C++17 we had to do some complicated logic to work with paths, this is solved with the `<filesystem> header, now present at REST.

Some methods to work with paths in REST relied on environment variables to work, such as `PWD`. This was causing some unexpected issues on my dev environment when the `PWD` env variable doesn't match the current working directory.

I have removed all apperances of environment variables when working on paths and I have rewritten the old methods to use `<filesystem>`. Now these methods are a single line and could probably be removed altogether so people start using the filesystem library, but they remain for now.

I removed the method `TRestTools::ReturnToPreviousDirectory` as it relied on environment variables for state management and it was only used once in the whole code base. I replaced its usage in restG4.

Related PR:

* https://github.com/rest-for-physics/restG4/pull/48/files

### DIFF
--- a/source/framework/core/src/TRestGDMLParser.cxx
+++ b/source/framework/core/src/TRestGDMLParser.cxx
@@ -84,12 +84,11 @@ void TRestGDMLParser::Load(const string& filename) {
 TGeoManager* TRestGDMLParser::CreateGeoManager() {
     // We must change to the gdml file directory, otherwise ROOT cannot load.
     if (!fOutputGdmlFilename.empty()) {
-        char originDirectory[256];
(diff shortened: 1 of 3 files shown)


juanangp (approved):

DavidDiezIb (approved):

jgalan (approved):
Just tested that I could launch `restG4` from a different directory. It was working as expected.

PR #230, opened 2022-05-26
Adding new library legacylib

juanangp:
juanangp Ok: 31 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/legacylib)

New `legacy` library has been added to the repository, check https://github.com/rest-for-physics/legacylib for more details.

The new  `legacy` library is meant to store deprecated libraries in order to keep backward compatibility of the data with the following conditions:
- A `legacy` class can be instanciated, in this case a warning in displayed.
- A `legacy` class doesn't implement any function e.g. "ProcessEvent", an error is displayed and the program exit in case.
- A `legacy` class is capable to print any metadata.
- All  `legacy` classes inherits from `TRestLegacyProcess`, which ensure the proper handling of legacy classes.

For the time being only `TRestRawZeroSuppresionProcess.h` is implemented as legacy class, however more classes can be added following this template. A base class and a template for `TRestLegacyMetadata` can be provided in case of need.

Also, when REST has not been compiled with all libraries, when trying to read a file that contains classes from the missing libraries, it would produce a seg.fault. This is fixed now by @jgalan

- Added method `REST_StringHelper::FirstToUpper`.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -67,3 +67,6 @@
 [submodule "projects/alphacamm"]
 	path = projects/alphacamm
 	url = git@lfna.unizar.es:gifna/alphacamm.git
+[submodule "source/libraries/legacy"]
+	path = source/libraries/legacy
+	url = https://github.com/rest-for-physics/geant4lib.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -69,4 +69,4 @@
 	url = git@lfna.unizar.es:gifna/alphacamm.git
 [submodule "source/libraries/legacy"]
 	path = source/libraries/legacy
-	url = https://github.com/rest-for-physics/geant4lib.git
+	url = https://github.com/rest-for-physics/legacylib.git
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -84,8 +84,8 @@ Build and Test:
     - cd ${CI_PROJECT_DIR}
     - python3 pull-submodules.py --force --dontask --latest:${CI_COMMIT_BRANCH}
     - mkdir ${CI_PROJECT_DIR}/build && cd ${CI_PROJECT_DIR}/build
-    - cmake ${CI_PROJECT_DIR}
-      -DTEST=ON -DREST_GARFIELD=OFF -DREST_G4=ON -DRESTLIB_DETECTOR=ON -DRESTLIB_RAW=ON -DRESTLIB_TRACK=ON
+    - cmake ${CI_PROJECT_DIR} -DTEST=ON -DREST_ALL_LIBS=ON
+      -DREST_GARFIELD=OFF -DREST_G4=ON -DREST_WELCOME=OFF
     - make -j2
     - ctest --output-on-failure -O ${CI_PROJECT_DIR}/build/Testing/summary.txt
 
@@ -103,7 +103,7 @@ Build and Install:
     - rm -rf ${CI_PROJECT_DIR}/build && mkdir ${CI_PROJECT_DIR}/build && cd ${CI_PROJECT_DIR}/build
     - rm -rf ${CI_PROJECT_DIR}/install
     - cmake ${CI_PROJECT_DIR} -DCMAKE_INSTALL_PREFIX=${CI_PROJECT_DIR}/install
-      -DREST_GARFIELD=ON -DREST_G4=ON -DRESTLIB_DETECTOR=ON -DRESTLIB_RAW=ON -DRESTLIB_TRACK=ON -DRESTLIB_AXION=ON -DREST_WELCOME=OFF
+      -DREST_ALL_LIBS=ON -DREST_GARFIELD=ON -DREST_G4=ON -DREST_WELCOME=OFF
     - make install -j2
 
   artifacts:
--- a/scripts/reformat-files.sh
+++ b/scripts/reformat-files.sh
@@ -46,6 +46,7 @@ if ! type -P $CLANG_FORMAT; then
   CLANG_FORMAT_DISABLED=True
 fi
 
+export XMLLINT_INDENT="    "
 if ! type -P $XML_LINT; then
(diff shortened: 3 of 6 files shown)


.gitmodules
```diff
@@ -67,3 +67,6 @@
 [submodule "projects/alphacamm"]
 	path = projects/alphacamm
 	url = git@lfna.unizar.es:gifna/alphacamm.git
+[submodule "source/libraries/legacy"]
+	path = source/libraries/legacy
+	url = https://github.com/rest-for-physics/geant4lib.git
```
jgalan, line 72: Be careful, this points to geant4lib!

juanangp: > Be careful, this points to geant4lib!

Such a big failure, I don't know how this end up like that. Thanks for pointing to it, now should be fine.

jgalan (approved):

lobis:
> Just realize that not all the libs are build by default, from https://github.com/rest-for-physics/framework/blob/master/.gitlab-ci.yml I can see:
> 
> ```
>  cmake ${CI_PROJECT_DIR} -DCMAKE_INSTALL_PREFIX=${CI_PROJECT_DIR}/install
>       -DREST_GARFIELD=ON -DREST_G4=ON -DRESTLIB_DETECTOR=ON -DRESTLIB_RAW=ON -DRESTLIB_TRACK=ON -DRESTLIB_AXION=ON -DREST_WELCOME=OFF
> ```
> 
> Which leds to: image
> 
> As you can see, some libraries such as `geant4` or `connectors` are not enabled at this stage of the build. However, I think they are enabled lated on due to some dependencies. I would suggest to enable all the libraries by default in the pipeline e.g.:
> 
> ```
>  cmake ${CI_PROJECT_DIR} -DCMAKE_INSTALL_PREFIX=${CI_PROJECT_DIR}/install -DREST_ALL_LIBS=ON -DREST_GARFIELD=ON -DREST_G4=ON -DREST_WELCOME=OFF
> ```
> 
> Same applies to the test, @lobis @jgalan let me know what do you think, I can do these changes and test in this PR.

I think its fine to add all libraries. Personally I like to have them all explicitly stated instead of having a single parameter for all of them, just in case we need to turn a single one off for some issue etc.

juanangp:
> I think its fine to add all libraries. Personally I like to have them all explicitly stated instead of having a single parameter for all of them, just in case we need to turn a single one off for some issue etc.

Well, in this case you can always do:
```
 cmake ${CI_PROJECT_DIR} -DCMAKE_INSTALL_PREFIX=${CI_PROJECT_DIR}/install -DREST_ALL_LIBS=ON -DREST_GARFIELD=ON -DREST_G4=ON -DREST_WELCOME=OFF -DRESTLIB_LEGACY=OFF
```
At least for me is more clear this way.

jgalan:
Just off-topic, but when reformating XML files, it adds 2-spaces, should not be 4-spaces as C-code?

Also it puts everything really compactified, I really like to allow at least 1-line space between definitions 

```
  </TRestSpiderMask>
  <TRestSpiderMask name="spider2" verboseLevel="info">
    <parameter name="maskRadius" value="20cm"/>
    <parameter name="offset" value="(0,0)cm"/>
    <parameter name="rotationAngle" value="0"/>
    <parameter name="armsWidth" value="5deg"/>
    <parameter name="armsSeparationAngle" value="30degrees"/>
    <parameter name="initialRadius" value="3cm"/>
  </TRestSpiderMask>
  <TRestSpiderMask name="spider3" verboseLevel="info">
    <parameter name="maskRadius" value="20cm"/>
    <parameter name="offset" value="(0,0)cm"/>
    <parameter name="rotationAngle" value="30deg"/>
    <parameter name="armsWidth" value="5deg"/>
    <parameter name="armsSeparationAngle" value="60degrees"/>
    <parameter name="initialRadius" value="6cm"/>
  </TRestSpiderMask>
</masks>
```

lobis:
> Just off-topic, but when reformating XML files, it adds 2-spaces, should not be 4-spaces as C-code?
> 
> Also it puts everything really compactified, I really like to allow at least 1-line space between definitions
> 
> ```
>   </TRestSpiderMask>
>   <TRestSpiderMask name="spider2" verboseLevel="info">
>     <parameter name="maskRadius" value="20cm"/>
>     <parameter name="offset" value="(0,0)cm"/>
>     <parameter name="rotationAngle" value="0"/>
>     <parameter name="armsWidth" value="5deg"/>
>     <parameter name="armsSeparationAngle" value="30degrees"/>
>     <parameter name="initialRadius" value="3cm"/>
>   </TRestSpiderMask>
>   <TRestSpiderMask name="spider3" verboseLevel="info">
>     <parameter name="maskRadius" value="20cm"/>
>     <parameter name="offset" value="(0,0)cm"/>
>     <parameter name="rotationAngle" value="30deg"/>
>     <parameter name="armsWidth" value="5deg"/>
>     <parameter name="armsSeparationAngle" value="60degrees"/>
>     <parameter name="initialRadius" value="6cm"/>
>   </TRestSpiderMask>
> </masks>
> ```

Yes I agree, I also prefer 4 spaces. I made a commit to this branch that should modify this (https://github.com/rest-for-physics/framework/commit/6caf746121aea39a985e65eaab0fbf972bad2eaa), can you please check it works?

Regarding the separation, I am not sure how to modify this. It seems that `xmllint` tool is really old and does not have many configuration settings, perhaps there is a better tool for this?

nkx111:
Talking about legacy, I guess the old V2.2 processes and events can be also added? If it is possible, then we can recover backward compatibility to v2.2.

juanangp:
> Talking about legacy, I guess the old V2.2 processes and events can be also added? If it is possible, then we can recover backward compatibility to v2.2.

Yes, this should be possible.

nkx111 (approved):

PR #232, opened 2022-05-30
Implementing read of child metadata elements automatically

jgalan:
jgalan Ok: 73 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/jgalan_child_metadata)

I am trying the implementation of an automatic read of child metadata definitions.

For example, if I have the following class definition:

```
    TRestMyClass::TRestMetadata {
              Double_t fMyParameter = 0;
            
              TRestOtherMdClass *fMdName = nullptr;
```

Then, I can create the following RML definition.

```
<TRestMyClass name="awesome_md" >
     <parameter name="myParameter" value="3.15" />

     <TRestOtherMdClass name="mdName">
            ....

     </TRestOtherMdClass>
</TRestMyClass>
```

I think it would be really interesting that we are able to construct metadata classes with sub-metadata classes, and initialize the corresponding metadata member.

What do you think? @rest-for-physics/core_dev 

I have been trying to do a first draft implementation but I am not very familiar with reflection, and not sure if it is that easy to achieve that.

The first application to test these new routines can be found at rest-for-physics/axionlib#25 

I also thought about the possibility of having a `std::vector <TRestOtherMdClass *>` so that it there are several definitions we place them in a vector.

I implemented `GetChildrenList` that returns a list of children with their respective name, then I attempt to create an instance pointing to the memory address. But for the moment no sucess :(

Any help is welcome!

### DIFF
--- a/source/framework/core/inc/TRestMetadata.h
+++ b/source/framework/core/inc/TRestMetadata.h
@@ -135,11 +135,17 @@ class TRestMetadata : public TNamed {
     virtual Int_t LoadSectionMetadata();
     /// To make settings from rml file. This method must be implemented in the derived class.
     virtual void InitFromConfigFile() {
+        RESTDebug << "TRestMetadata::InitFromConfigFile" << RESTendl;
         std::map<std::string, std::string> parameters = GetParametersList();
 
         for (auto& p : parameters) p.second = ReplaceMathematicalExpressions(p.second);
 
         ReadParametersList(parameters);
+
+        /// For the moment this method will attempt to instantiate, just like that for testing
+        std::vector<std::pair<std::string, std::string>> mdChildren = GetChildrenList();
+
+        //       for (const auto x : mdChildren) std::cout << x.first << " : " << x.second << std::endl;
     }
 
     /// Method called after the object is retrieved from root file.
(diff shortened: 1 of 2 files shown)


nkx111:
I am not sure `TRestReflector::GetDataMember()` will give you child metadata classes. This function only gives datamember without `//!` notation because only those can be identified by ROOT streamer. Besides, if the data member is a pointer, it may not totally owned by the container class. Enabling autamatic reading may cause conflicts.

jgalan:
Could you have a look to `InstantiateChild` method? I am almost there, but now I should assign the pointer `TRestMetadata *md` to `datamember` as it is done with others such as `*(double *)` is not the way?

jgalan:
Ok, then I have just created a helper method inside `TRestMetadata` so that we avoid `TiXml` code spreading in child classes.

See PR #233 

We can use that method to make the task easier. I.e. I use as:

```
    if (fMask) {
        delete fMask;
        fMask = nullptr;
    }
    fMask = (TRestPatternMask*)this->GetChildMetadata("Mask", "mask");
```

nkx111:
I got you wrong, I was thinking reading child metadata elements from class definition and find the rml section accordingly. Now the inversed approch seems to be much better. We initialize metadata objects from rml, and set class members by hand. Looks nice! We can use this method also in TRestProcessRunner and TRestRun.

jgalan:
This is probably obsolete, it was not achieved but simplified by implementing the method `TRestMetadata::InstantiateChildMetadata`

PR #233, opened 2022-05-30
TRestMetadata::InstantiateChildMetadata method added

jgalan:
jgalan Ok: 94 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/jgalan_metadata_child)

Simply adding a method that will help to retrieve child TRestMetadata XML elements.

This will allow to keep `TinyXml` into the domain of `TRestMetadata`.

We can use that method to make the task of reading a child metadata element easier. I.e. I use inside `TRestAxionXrayWindow` as:

```
    if (fMask) {
        delete fMask;
        fMask = nullptr;
    }
    fMask = (TRestPatternMask*)this->GetChildMetadata("Mask", "mask");
```


source/framework/core/src/TRestMetadata.cxx
```diff
@@ -656,6 +656,54 @@ Int_t TRestMetadata::LoadConfigFromBuffer() {
...
+/// belongs to the geant4 library.
+///
+/// - *name*: It can be specified a specific given name.
+///
+/// Otherwise, the first child section that satisfies that it starts by `TRest` will be
+/// returned.
+///
+/// If no child element is found `nullptr` will be returned.
+///
+TRestMetadata* TRestMetadata::GetChildMetadata(std::string pattern, std::string name) {
```
nkx111, line 683: Maybe `InitializeChildMetadata()` could be a better name?

jgalan: Perhaps `InstantiateChildMetadata`?

source/framework/core/src/TRestMetadata.cxx
```diff
@@ -656,6 +656,54 @@ Int_t TRestMetadata::LoadConfigFromBuffer() {
...
+    auto paraele = fElement->FirstChildElement();
+    while (paraele != nullptr) {
+        std::string xmlChild = paraele->Value();
+        if (xmlChild.find("TRest") == 0) {
+            if (pattern == "" || xmlChild.find(pattern) != string::npos) {
+                if (name != "" && name == (string)paraele->Attribute("name")) {
```
juanangp, line 689:
```suggestion
            if (pattern.empty() || xmlChild.find(pattern) != string::npos) {
                if (!name.empty() && name == (string)paraele->Attribute("name")) {
```

source/framework/core/src/TRestMetadata.cxx
nkx111: Yes!

juanangp (approved):

nkx111 (approved):

Issue #238, opened 2022-06-02
Update pipeline image to use latest version of dependencies


lobis:
Currently the image is using a significantly old version of Geant4, 10.4.3 for compatibility reasons.

REST is now compatible with latest Geant4 version, but probably some things such as validation scripts should be updated when we increase version.

PR #239, opened 2022-06-02
Use latest dependencies in pipeline image

lobis:
lobis Large: 1665 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/238-update-pipeline-image-to-use-latest-version-of-dependencies)

Closes https://github.com/rest-for-physics/framework/issues/238.

The latest official root image uses Ubuntu 22.04 as base so I think we should use it as a base too, instead of Ubuntu 20.04. There may be some problems due to this such as cmake version changes etc.

In this PR I renamed a few files in `pipeline/pandaxiii_MC`. I noticed one of the files used as reference to check simulation (originally named `validation.root` I think) was taking too much space, around 25MB but only its first 10 events were being used. I reduced it in size saving around 23MB of space, but we would need to remove big blobs from the history doing a force push in order to reduce the repository size.

Related PR:

* https://github.com/rest-for-physics/restG4/pull/51

Part of this PR was merged in https://github.com/rest-for-physics/framework/pull/248


lobis:
> Now that it seems you are updating the docker image. It could be added `-DSQL=ON` as compilation option? This will require installing MySQL dev libraries (if not already available).

Sure, I can add it. I think these libraries are installed. I found no issue when enabling `-DSQL=ON` but no new target is being built, however I see:

```
===========================================================
System is: Linux
Packages: restG4 
Libraries: RestFramework;RestConnectors;RestDetector;RestGeant4;RestRaw
Features: Eve;SQL
===========================================================
```

Maybe I need `-DREST_SQL=ON`?

nkx111:
Actually it feels bad to me changing geant4 version in the pipeline. This breaks consistency. Is it necessary to do that? If the latest version is needed, for example, to test some new code feature, I think it is better to insert an individual stage in the pipeline with a new docker image.

As for tolerance, I think we should not use it too much in the validation file. If some commits are being pushed influencing simulation result, by using tolerance we won't be able to notice them from pipeline. We shall keep several jobs validated with strict values.

lobis:
> Actually it feels bad to me changing geant4 version in the pipeline. This breaks consistency. Is it necessary to do that? If the latest version is needed, for example, to test some new code feature, I think it is better to insert an individual stage in the pipeline with a new docker image.
> 
> As for tolerance, I think we should not use it too much in the validation file. If some commits are being pushed influencing simulation result, by using tolerance we won't be able to notice them from pipeline. We shall keep several jobs validated with strict values.

Personally I think its good that the code is tested with a fairly recent version of Geant4. The motivator for this change was that recently the official docker image of root was changed to be based on Ubuntu 22.04 instead of 20.04 and I couldn't install Geant4 10.4.3 on that version due to missing dependencies. The v10.4.3 version of Geant4 we are using is 3.5-4 years old, I think it may be time to update it on the pipeline now that REST also works with newer versions of Geant4.

Regarding tolerance I get your point but I think its better to have a validation code that can be run from different Geant4 versions and still work (unless big changes are present). If we use frozen values to compare and validate, changing Geant4 version will make our validation fail due to difference in random generation (as you pointed out in https://github.com/rest-for-physics/restG4/pull/51#discussion_r891374030), not due to a difference in the underlying physics. That is why I introduced tolerance and also made some examples validate on event level aggregate observables instead of single event observables.

Also I think that the pipeline / simulation examples should be moved to a test suite instead, and run the tests on the pipeline. This would allow the user to quickly run the validation on their local machine via for example `ctest`. If we limit ourselves to a single Geant4 version, tests will fail unless user uses that exact version.

I have kept some validation such as example `restG4/08.Alphas` to partially check exact values (such as the counts of each simulation) so that hopefully we can still detect small changes in simulation output. This example is run on the `framework` pipeline.

Please let me know what do you think @jgalan @juanangp about this PR. Right now its finished and ready to merge unless you request some other change or point out any issue.

jgalan:
We should keep REST compatible with oldest versions too. It is important that we can check the results coming out from 10.4 are the same as the ones we get with 10.7, 11, etc.

In that sense, the docker images are pulled at each stage, so I dont know if it is possible to run both?

I am always afraid that the efforts we dedicated to the validation might have leaks in this kind of updates, and get lost. I would try to keep the old pipelines running forever. I dont undertand why this would not be possible.

Sorry, I may not follow 100% but these are my partial thoughts.

You know, I am always in favor of progress but respecting the past, I am always betting for keeping REST compatible with oldest system as possible, preserving the existing pipelines will guarantee that. We will be able to install REST without major difficulties in clusters with poor system maintenance.

lobis:
> We should keep REST compatible with oldest versions too. It is important that we can check the results coming out from 10.4 are the same as the ones we get with 10.7, 11, etc.
> 
> In that sense, the docker images are pulled at each stage, so I dont know if it is possible to run both?
> 
> I am always afraid that the efforts we dedicated to the validation might have leaks in this kind of updates, and get lost. I would try to keep the old pipelines running forever. I dont undertand why this would not be possible.
> 
> Sorry, I may not follow 100% but these are my partial thoughts.
> 
> You know, I am always in favor of progress but respecting the past, I am always betting for keeping REST compatible with oldest system as possible, preserving the existing pipelines will guarantee that. We will be able to install REST without major difficulties in clusters with poor system maintenance.

I agree.

We need to check REST is able to be build with a reference Geant4 version, such as 10.4.3. This check was introduced in https://github.com/rest-for-physics/restG4/pull/52. Using multiple images in the pipeline is not a problem.

However I don't think its practical to run the whole pipeline for different versions of Geant4, since there are changes we need to make to the validation scripts that come from the fact that the random numbers are not the same accross versions.

I think it would be enough to have a reference simulation file created with a reference Geant4 version and we just check that this file is reproduced correctly using the reference Geant4 version.

juanangp:
> I think this can be merged to fix pipeline @juanangp @jgalan. Gitlab pipeline has been updated to latest image version but the GitHub pipeline now also verifies examples run on old geant4 so I think we are covered.

Apologizes, I did a small mesh, I will accept as soon as I fix that. Btw, do you know why `alphatrack` validation is so ridiculous slow? It was not happening with previous `geant4` version

juanangp (approved):

jgalan:
I think ROOT validation files should never be modified/replaced. Perhaps adding new versions?

jgalan:
Probably ROOT validation files could be placed at an independent repository, and the files could be made public in a http location. Such as it is done with https://sultan.unizar.es/axionlib-data/ and the repository https://github.com/rest-for-physics/axionlib-data

lobis:
> Probably ROOT validation files could be placed at an independent repository, and the files could be made public in a http location. Such as it is done with https://sultan.unizar.es/axionlib-data/ and the repository https://github.com/rest-for-physics/axionlib-data.
> 
> The http address is updated nightly in sultan through the `rest` user.
> 
> Then, the validation pipeline could just get the validation file needed using `wget`.

I agree but I think for the time being this PR could be merged.

jgalan (commented):
I approve just to fix the pipelines, by this PR with 2000 lines changes, renaming of validation ROOT files, probably replacement. It is too much, thus imposible to review.

jgalan (approved):

PR #241, opened 2022-06-06
Using `whoami` when USER is not defined instead of default user

lobis:
lobis Ok: 12 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/lobis-default-user)

Currently if the `USER` environment variable is not set, `defaultUser` is set as `USER`.

This PR makes use of `whoami` (which also exists on windows apparently) to get the current username, which is not always present in the environtment variable. The `USER` env variable will still take precendence.

If for some reason `whoami` is not found, it will still use `defaultUser` and it won't raise an exception.

### DIFF
--- a/source/framework/core/src/startup.cpp
+++ b/source/framework/core/src/startup.cpp
@@ -27,7 +27,6 @@ map<string, RESTVirtualConverter*> RESTConverterMethodBase = {};
 struct __REST_CONST_INIT {
    public:
     __REST_CONST_INIT() {
-
         REST_COMMIT = TRestTools::Execute("rest-config --commit");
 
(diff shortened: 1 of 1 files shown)


lobis:
pipeline is failing due to some other reason. This works locally.

nkx111:
> pipeline is failing due to some other reason. This works locally.

I mistakenly merged the PR https://github.com/rest-for-physics/restG4/pull/49#. It should be merged only after https://github.com/rest-for-physics/framework/pull/231 is merged. Shall we revert it or merge the later?

cmargalejo (approved):

PR #244, opened 2022-06-12
Nkx111 patch 4

nkx111:
nkx111 Medium: 166 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/nkx111-patch-4) [](https://github.com/rest-for-physics/framework/commits/nkx111-patch-4)

Some bug fix and new features:
1. Fix bugs when opening files with unknown event types(e.g., v2.2 file). Previsuly in this case the input event will be initialized with data members, but the method cannot be called, and causes seg.fault. 
2. added method: TRestEventProcess::GetParallelDataMembers()
3. restRoot macro loading now also working on windows.
4. added validation pipeline to test open v2.2 files.
5. fixed Console::kbhit() on windows.


source/bin/restRoot.cxx
```diff
@@ -77,11 +77,11 @@ int main(int argc, char* argv[]) {
     gROOT->ProcessLine("#include <TRestPhysics.h>");
     if (loadMacros) {
         if (!silent) printf("= Loading macros ...\n");
-        auto a = TRestTools::Execute(
-            "find $REST_PATH/macros | grep REST_[^/]*.C | grep -v \"swo\" | grep -v "
-            "\"CMakeLists\" | grep -v \"swp\"  | grep -v \"svn\"");
-        auto b = Split(a, "\n");
-        for (auto c : b) {
+        auto a = TRestTools::GetFilesMatchingPattern(REST_PATH + "/macros/*REST_*.C");
+        for (auto c : a) {
```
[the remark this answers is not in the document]
nkx111: fixed.

source/framework/core/inc/TRestHits.h
```diff
@@ -173,10 +173,11 @@ class TRestHits : public TObject {
     Double_t GetMinimumHitEnergy() const;
     Double_t GetMeanHitEnergy() const;
 
-    Double_t GetEnergyIntegral() const;
+    Double_t CalculateTotalDepositedEnergy();
```
lobis, line 176: Personally I think this method name is misleading, I think all methods that return a result and do not alter anything (are marked `const`) should start by `Get`, in this case maybe it should be `GetTotalDepositedEnergy`. This method should be `const`.

In my opinion a method named like this would be expected to update some internal member with some expensive to calculate magnitude which could be later accessed via a getter, and this calculation is not directly implemented in the getter because its expensive.

source/framework/tools/src/TRestTools.cxx
```diff
@@ -130,7 +130,7 @@ void TRestTools::LoadRESTLibrary(bool silent) {
             if (excluded) {
                 continue;
             }
-            fileList.emplace_back(it.path());
+            fileList.emplace_back(it.path().string());
```
lobis, line 133: It is not necessary to add `.string()` since it will automatically convert it, but this is a matter of opinion on which way it looks more clean.

source/framework/tools/inc/TRestReflector.h
```diff
@@ -242,8 +242,13 @@ inline TClass* GetClassQuick(std::string type) {
         return iter->second;
     } else {
         TClass* cl = TClass::GetClass(type.c_str());
-        RESTListOfClasses_typename[type] = cl;
-        return cl;
+        if (cl != NULL && cl->HasDictionary()) {
```
lobis, line 245:
```suggestion
        if (cl != nullptr && cl->HasDictionary()) {
```

source/framework/core/src/TRestTask.cxx
```diff
@@ -196,7 +204,7 @@ void TRestTask::RunTask(TRestManager* mgr) {
                 } else {
                     string type = meta->ClassName();
                     string cmd = Form("%s* %s = (%s*)%s;", type.c_str(), fInvokeObject.c_str(), type.c_str(),
-                                      ToString(meta).c_str());
+                                      (PTR_ADDR_PREFIX+ToString(meta)).c_str());
```
lobis, line 207:
```suggestion
                                      (PTR_ADDR_PREFIX + ToString(meta)).c_str());
```

.gitlab-ci.yml
```diff
@@ -114,19 +114,26 @@ Build and Install:
...
     - restRoot -b -q
 
 List REST Macros:
-  stage: install
+  stage: simpleRun
   script:
     - source ${CI_PROJECT_DIR}/install/thisREST.sh
     - restManager ListMacros
 
+Open v2.2 Files:
```
lobis, line 130: We are in the process of migrating to GitHub Actions so probably we should not add new steps into the GitLab pipeline, instead we should add them into the GitHub.

jgalan (approved):
Hi @nkx, perhaps we could create a dedicated data repository, where we place data files such as the one you are adding `v2.2.30_hits.root`?

This would be something similar to https://github.com/rest-for-physics/axionlib-data/tree/36629372bcffddd6f95eb12a565c1b15475a64be

That we later make http accessible, and use at the pipelines using `wget`.

jgalan:
What the `GetParallelDataMembers` is doing?

source/framework/core/inc/TRestHits.h
juanangp: I agree with you, can be renamed but should be stated that is a `Getter` and should be `const`

source/framework/core/inc/TRestEventProcess.h
```diff
@@ -108,6 +108,36 @@ class TRestEventProcess : public TRestMetadata {
...
+    /// 
+    /// Because all the parallel processes are in same data structure, this method directly 
+    /// gets the data members by accessing certain offset.
+    /// 
+    /// Usage: (suppose we defined fHist as data member)
+    /// auto hists = this->GetParallelObjects(&fHist);
+    /// for (auto h : hists) { fHist->Add(h); } 
+    /// 
+    template <class T>
+    std::vector<T> GetParallelDataMembers(T* member_of_process) {
```
juanangp, line 125: Where is this template used?

source/framework/core/inc/TRestEventProcess.h
```diff
@@ -229,4 +259,9 @@ class TRestEventProcess : public TRestMetadata {
 
     ClassDefOverride(TRestEventProcess, 3);  // Base class for a REST process
 };
+
+#define _ApplyCut(evt)        \
```
juanangp, line 263: Personally, I am not fan of macros. Also, I cannot see where this macro is used.

source/framework/core/inc/TRestHits.h
juanangp: In fact is changing a data member `fTotEnergy`, so you can do
```
void CalculateTotalDepositedEnergy(){fTotEnergy = GetEnergyIntegral(); }
```

nkx111: > In fact is changing a data member fTotEnergy, so you can do

That's what I want. We lack a method that updates `fTotEnergy` data member, in case some processes modifies hit point directly.  I can make the method void in this way.

source/framework/tools/src/TRestTools.cxx
nkx111: If not add `.string()` there will be problem compiling under windows.

.gitlab-ci.yml
nkx111: I am not familar with GitHub Actions, I need to study a little.

nkx111:
> Hi @nkx, perhaps we could create a dedicated data repository, where we place data files such as the one you are adding `v2.2.30_hits.root`?
> 
> This would be something similar to https://github.com/rest-for-physics/axionlib-data/tree/36629372bcffddd6f95eb12a565c1b15475a64be
> 
> That we later make http accessible, and use at the pipelines using `wget`.

It sounds interesting. We also have several root files used in pipeline for process input, and generator files for NLDBD simulation. Those files are much larger than others.

source/framework/tools/src/TRestTools.cxx
lobis: > If not add `.string()` there will be problem compiling under windows.

Oh, I didn't know, good to know.

.gitlab-ci.yml
nkx111: I have migrated the file opening tests to GitHub Actions

source/framework/core/inc/TRestEventProcess.h
nkx111: It could used in for example in `TRestRawSignalAnalysisProcess`, to replace the cut code. 

`if (ApplyCut()) return nullptr; return fSignalEvent;` --> `_ApplyCut(fSignalEvent)`

I think it is more clear to use this macro.

juanangp (approved):

lobis (approved):

PR #251, opened 2022-06-15
TRestAnalysisPlot. Now addFile supports filename patterns

jgalan:
jgalan Ok: 18 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/jgalan_analysis_plots) [](https://github.com/rest-for-physics/framework/commits/jgalan_analysis_plots)

This is to allow `addFile` definitions inside `TRestAnalysisPlot` as `<addFile name="/path/to/file/Run*.root" />`

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -78,17 +78,17 @@ Validate Code:
     - python3 pull-submodules.py --force --dontask --latest:${CI_COMMIT_BRANCH} --exclude:iaxo,detector-template
     - python3 pipeline/validateProcesses.py source/libraries/
 
-Build and Test:
-  stage: test
-  script:
-    #- cd ${CI_PROJECT_DIR}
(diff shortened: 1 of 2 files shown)


KonradAltenmueller (approved):

lobis:
> Any idea why pipeline is failing? Seems connected to restG4 @lobis?

No idea, the GitHub pipeline passes so I would just merge it if it's urgent. I will look at it, but probably we should move to using GitHub only to avoid these problems, maybe @juanangp has some thoughts?

cmargalejo (approved):

juanangp:
> itlab-ci.yml Removing build tests from pipeline

Should be fixed now, but we have to check

jgalan:
In principle if those build-tests are already at GitHub, it could reduce the workload at the Gitlab pipeline.

PR #253, opened 2022-06-16
Added global TRestRun store

lobis:
lobis Medium: 153 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/lobis-run-global-store) [](https://github.com/rest-for-physics/framework/commits/lobis-run-global-store)

This PR adds a global store which provides access to all currently existing `TRestRun` objects from a static method. This is added at 6e769158278e48515fe1508f6d2c6af4e49f155f. This PR also has some small non-functional changes to the `TRestRun` class.

A test to check this is also added in this same PR.

How it works:

* The constructor of `TRestRun` will add a pointer to itself to the global store (which is static vector).
* The destructor of `TRestRun` will remove the pointers from the vector. This means that by construction all pointers inside the global store cannot be null, since once they are deleted, they get removed automatically.

Why this change was introduced:

This change does not add any overhead and provides a convinient way to access the current run without having to do injection. For example a `TRestGeant4Track` which is stored in a REST file may use this to access its `TRestRun` and from it get access to its corresponding `TRestGeant4Metadata` which could be used to get geometry information on the `PrintTrack` method, allowing to print volume names instead of ids which is what we are limited to at the moment.

### DIFF
--- a/source/framework/core/inc/TRestRun.h
+++ b/source/framework/core/inc/TRestRun.h
@@ -101,7 +101,8 @@ class TRestRun : public TRestMetadata {
     }
 
     TString FormFormat(const TString& FilenameFormat);
-    TFile* MergeToOutputFile(std::vector<std::string> filefullnames, std::string outputfilename = "");
+    TFile* MergeToOutputFile(const std::vector<std::string>& filefullnames,
+                             const std::string& outputfilename = "");
     TFile* FormOutputFile();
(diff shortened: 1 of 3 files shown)


lobis:
This approach has been deprecated in favour of having a reference to the corresponding `TRestRun` stored inside `TRestEvent` which will be updated on `TRestRun::GetEntry`.

PR #254, opened 2022-06-17
Gauss hit zeroes fit

cmargalejo:
cmargalejo Large: 559 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/gaussHitZeroesFit) [](https://github.com/rest-for-physics/framework/commits/gaussHitZeroesFit)

This pull request aims to solve some issues that were found with the observables GaussSigmaX, GaussSigmaY and GaussSigmaZ, which are computed by fitting a gaussian to each event.

Some of the events were found to have sigmas that were too large. After a deep study, the hypothesis is that the fit fails especially for small events that have too few hits.  Not seeing a hit doens't necesarily mean there is no energy on that channel, but that it is below the threshold. Thus, the proposed  solution is to add one hit to the left and one hit to the right, at a value 0+/-error.

The main changes are:
-Adding two hits to each hits event.
-Define the energy of these hits as 0 +/- 70 ADC. This error is defined based on typical for the micromegas detectors.

To illustrate the changes, the image below shows the same event before the changes in the code in the left, and the new fit in the right.

image


jgalan:
Hi @cmargalejo , the problem with the pipeline is surely related to the fact that your branch is not updated with master.

You should do locally:

```
git checkout master
git pull
git checkout gaussHitZeroesFit
git merge master
git push
```

So that this PR gets updated

source/framework/core/src/TRestHits.cxx
```diff
@@ -545,103 +546,183 @@ void TRestHits::GetBoundaries(std::vector<double>& dist, double& max, double& mi
 Double_t TRestHits::GetGaussSigmaX() {
```
jgalan, line 546: It would be good to add a description to this method. Perhaps reusing some text from the PR description. I see this class is not following the standard headers. I will just quickly updated that

jgalan:
> -Define the energy of these hits as 0 +/- 70 ADC. This error is defined based on typical for the micromegas detectors.

What it happens if you set the new hits on the sides to equal zero? If you add the new hits, then the energy of the event will be affected, or the hits are added just for the gaussian fit, then removed for further processing?

jgalan:
Perhaps it would be interesting to add the image you attached to the PR description to the documentation.

jgalan (approved):

cmargalejo:
> > -Define the energy of these hits as 0 +/- 70 ADC. This error is defined based on typical for the micromegas detectors.
> 
> What it happens if you set the new hits on the sides to equal zero? If you add the new hits, then the energy of the event will be affected, or the hits are added just for the gaussian fit, then removed for further processing?

Yes, the hits are only added to a TGraph to perform the fit, so they don't go into any further processing. Still, their energy is 0.

source/framework/core/src/TRestHits.cxx
cmargalejo: Description added in [154df75](https://github.com/rest-for-physics/framework/pull/254/commits/154df759bffd7b34b65165775201da4290d087fb)

cmargalejo:
I cannot compile REST from this point, as I get errors in `TRestTrackLinearizationProcess.cxx`. Does it happen to you?

KonradAltenmueller (approved):

source/framework/core/src/TRestHits.cxx
```diff
@@ -541,109 +759,204 @@ void TRestHits::GetBoundaries(std::vector<double>& dist, double& max, double& mi
...
+            x[0] = xMin - 0.5;
+            x[h] = xMax + 0.5;
+            y[0] = 0.0;
+            y[h] = 0.0;
+            ex[0] = 0.0;
+            ex[h] = 0.0;
+            ey[0] = 70.0;
+            ey[h] = 70.0;
+        }
+        TGraphErrors* grX = new TGraphErrors(nElems, &x[0], &y[0], &ex[0], &ey[0]);
```
nkx111, line 808: You should delete the newed variables.

source/framework/core/src/TRestHits.cxx
```diff
@@ -541,109 +759,204 @@ void TRestHits::GetBoundaries(std::vector<double>& dist, double& max, double& mi
...
+        fit->SetParameter(2, sigma);
         grX->Fit(fit, "QNB");  // Q = quiet, no info in screen; N = no plot; B = no automatic start
-                               // parmaeters; R = Use the Range specified in the function range
+                               // parameters; R = Use the Range specified in the function range
 
         gausSigmaX = fit->GetParameter(2);
     }
 
-    return gausSigmaX;
+    return abs(gausSigmaX);
```
nkx111, line 827: I suggest to add some checks if the fit result is negative or too large. In these cases we can return -1 to represent internal error.

nkx111 (changes requested):

source/framework/core/src/TRestHits.cxx
cmargalejo: I don't think a negative value is an error. It is a valid value. See https://root.cern.ch/doc/master/TMath_8cxx_source.html#l00448

For values that are too large, I'm trying the code below but it doesn't compile and I don't know why. Any suggestion?

```C++
        TF1* fit = new TF1("", "gaus");
        fit->SetParameter(0, maxY);
        fit->SetParameter(1, maxX);
        fit->SetParameter(2, sigma);
        TFitResultPtr fitResult = grX->Fit(fit, "QNBS");  // Q = quiet, no info in screen; N = no plot; B = no automatic start
                               // parameters; R = Use the Range specified in the function range; S = save and return the fit result.
        if(fitResult->isValid()){
            gausSigmaX = fit->GetParameter(2);
        }
        else{
            return -1.0; // the fit failed, return -1 to indicate failure
        }
```
The error I get is the following:

```
Consolidate compiler generated dependencies of target RestFramework
[  1%] Building CXX object source/framework/CMakeFiles/RestFramework.dir/core/src/TRestHits.cxx.o
/home/cristina/GitHub/rest/source/framework/core/src/TRestHits.cxx: In member function ‘Double_t TRestHits::GetGaussSigmaX()’:
/home/cristina/GitHub/rest/source/framework/core/src/TRestHits.cxx:825:21: error: invalid use of incomplete type ‘class TFitResult’
         if(fitResult->isValid()){
                     ^~
In file included from /programas/root/6.24.02/include/TGraph.h:30,
                 from /programas/root/6.24.02/include/TGraphErrors.h:24,
                 from /home/cristina/GitHub/rest/source/framework/core/inc/TRestHits.h:30,
                 from /home/cristina/GitHub/rest/source/framework/core/src/TRestHits.cxx:76:
/programas/root/6.24.02/include/TFitResultPtr.h:30:7: note: forward declaration of ‘class TFitResult’
 class TFitResult;
       ^~~~~~~~~~
make[2]: *** [source/framework/CMakeFiles/RestFramework.dir/build.make:734: source/framework/CMakeFiles/RestFramework.dir/core/src/TRestHits.cxx.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:240: source/framework/CMakeFiles/RestFramework.dir/all] Error 2
make: *** [Makefile:136: all] Error 2
cristina@sultan:~/GitHub/rest/build$

```

source/framework/core/src/TRestHits.cxx
cmargalejo: You are right. Done here: https://github.com/rest-for-physics/framework/pull/254/commits/d6b6d5687292be6a6b2d43f804c37f126668aa8e

source/framework/core/src/TRestHits.cxx
nkx111: Anyway it is just an suggestion. If you don't return -1 there is no problem for me.

nkx111 (approved):

PR #260, opened 2022-06-30
Added a pointer to TRestRun in TRestEvent

lobis:
lobis Ok: 19 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/lobis-trestrun-event-injection) [](https://github.com/rest-for-physics/framework/commits/lobis-trestrun-event-injection)

This PR adds a pointer to `TRestRun` to `TRestEvent` which is initialized when retrieving the event via `TRestRun::GetEntry`. This allows `TRestEvent` to access its corresponding metadata which can be used to improve print methods output for example.

https://github.com/rest-for-physics/geant4lib/pull/56 makes use of this PR.

### DIFF
--- a/source/framework/core/inc/TRestEvent.h
+++ b/source/framework/core/inc/TRestEvent.h
@@ -115,6 +115,6 @@ class TRestEvent : public TObject {
     // Destructor
     virtual ~TRestEvent();
 
-    ClassDef(TRestEvent, 2);
+    ClassDef(TRestEvent, 1);
 };
 #endif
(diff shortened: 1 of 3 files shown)


lobis:
@nkx111 do you know why this simple change (67fcdeb0919887303bae547288550cc4d0877652) makes the pipeline fail only at PandaX-III example?

juanangp (approved):

jgalan:
I think there is no problem to merge this, but I think it would be good if @nkx111 comments on it, since he implemented something similar inside `TRestEventProcess`. @nkx111  Do you see any issue with the present implementation?

Probably, then, it could be interesting that `restRoot` would initialize the events using the corresponding `TRestRun` instance.

jgalan (approved):

PR #263, opened 2022-07-11
Minor updates

lobis:
lobis Ok: 88 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/lobis-minor-updates) [](https://github.com/rest-for-physics/framework/commits/lobis-minor-updates)

In this PR I introduce small changes motivated by other PRs.

* Using `std::filesystem::weakly_canonical` to save the output file path inside `TRestRun` after "forming". Saving the path as a canonical absolute path helps solve a bug showing the wrong path at the end of simulation when launching from the IDE. Currently paths are being stored with things such as "./", these things are now resolved before storing and the path is absolute.
* `TRestHits` no longer inherits from `TObject` as this is not necessary unless we want to store the hits directly in the root file, but we always save the class containing the hits, such as a `TRestEvent`. This reduces the members of the class which reduces the visual clutter when using the root explorer. Also `TRestHits::RemoveHits` is now virtual so it can be overriden in other classes.
* All data members of `TRestHits` such as `fNHits` have been made protected instead of public and should be accessed by the corresponding getter which already existed before this PR.

Related PR:
* https://github.com/rest-for-physics/detectorlib/pull/55
* https://github.com/rest-for-physics/tracklib/pull/21
* https://github.com/rest-for-physics/connectorslib/pull/18


source/framework/core/inc/TRestHits.h
```diff
@@ -21,40 +21,32 @@
...
-#include "TCanvas.h"
-#include "TMath.h"
-#include "TMatrixD.h"
-#include "TObject.h"
-
 enum REST_HitType { unknown = -1, X = 2, Y = 3, Z = 5, XY = 6, XZ = 10, YZ = 15, XYZ = 30 };
-//! It let save an event as a set of punctual deposition.
+//! Allows saving an event as a set of punctual deposition.
 //! It saves a 3-coordinate position and an energy for each punctual deposition.
-class TRestHits : public TObject {
```
juanangp, line 42: Do you we have to increase the library version because we remove the inheritance from `TObject`? I am not sure of the backward compatibility in this case...

lobis: I was thinking no but I am not sure to be honest. I will increase it just in case.

juanangp (approved):

source/framework/core/src/TRestRun.cxx
```diff
@@ -1021,17 +1011,19 @@ TFile* TRestRun::MergeToOutputFile(vector<string> filenames, string outputfilena
...
     CloseFile();
+
     fOutputFileName = FormFormat(fOutputFileName);
+    // remove unwanted "./" etc. from the path while resolving them
+    fOutputFileName = std::filesystem::weakly_canonical(fOutputFileName.Data());
+
     fOutputFile = new TFile(fOutputFileName, "recreate");
     fAnalysisTree = new TRestAnalysisTree("AnalysisTree", "AnalysisTree");
     fEventTree = new TTree("EventTree", "EventTree");
-    // fAnalysisTree->CreateBranches();
```
jgalan, line 1033: Why this is not needed anymore? Perhaps @nkx has some comments here?

jgalan (approved):

source/framework/core/src/TRestRun.cxx
lobis: > Why this is not needed anymore? Perhaps @nkx has some comments here?

No idea, I will @nkx111 reply.

jgalan (approved):

PR #276, opened 2022-07-29
Adding MPFR library at the framework level

jgalan:
jgalan Large: 3120 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/event_processing) [](https://github.com/rest-for-physics/framework/commits/event_processing)

This PR simply brings the MPFR library to the framework level and not only for axionlib.

MPFR library allows calculations with any required number precision.

I will soon use it with a new class `TRestComplex` to do calculations with exact precision, but for the moment this change is required by the recent rest-for-physics/axionlib#12 PR.


jgalan:
Plz @lobis could you check it is the appropriate way to introduce this inside CMakeLists?

ruzarmen (approved):

lobis:
> Plz @lobis could you check it is the appropriate way to introduce this inside CMakeLists?

I made some modifications.

Instead of throwing an error when not found I set it to warning, I don't think we want to give an error when this library is not found right?

The previous implementation relied on the `MPFR_PATH` variable to exist, otherwise the call to `link_directories` would not work even if the library was in your system. Now it should work in more cases but I have not tested it.

In my machine using the pipeline docker image I don't need to specify this `MPFR_PATH` variable for it to work.

Maybe the `link_directories` step can be skipped, not sure.

jgalan:
Great addons @lobis, lets then approve and merge the changes?

lobis (approved):

source/framework/CMakeLists.txt
```diff
@@ -21,6 +21,19 @@ if (NOT ${CURL_LIB} STREQUAL "CURL_LIB-NOTFOUND")
...
+find_library(MPFR_LIB mpfr HINTS ${MPFR_PATH}) # path to .so file
+
+if (NOT MPFR_INCLUDE MATCHES "MPFR_INCLUDE-NOTFOUND" AND NOT MPFR_LIB MATCHES "MPFR_LIB-NOTFOUND")
+    # MPFR is found
+    add_compile_definitions(USE_MPFR)
+    set(external_include_dirs ${external_include_dirs} ${MPFR_INCLUDE})
+    set(external_libs "${external_libs};-lmpfr")
+    link_directories(${MPFR_LIB})
+else ()
+    message(WARNING "MPFR library was not found. Adding the location to the installation path may solve this problem.\n Use -DMPFR_PATH=/path/to/mpfr/")
```
nkx111, line 34: We should add some flags if this library is not linked, then one can type `rest-config --flags` to see if he can compile a MPFR-dependent REST library(like axionlib) based on this installation of REST

PR #277, opened 2022-07-29
Updated TRestMetadata::GetFieldValue to allow for default value

lobis:
Allowing for default value

### DIFF
--- a/source/framework/core/inc/TRestMetadata.h
+++ b/source/framework/core/inc/TRestMetadata.h
@@ -91,7 +91,8 @@ class TRestMetadata : public TNamed {
 
    protected:
     // new xml utilities
-    std::string GetFieldValue(std::string parName, TiXmlElement* e);
+    std::string GetFieldValue(std::string parName, TiXmlElement* e,
+                              std::string defaultValue = "Not defined") const;
     std::string GetParameter(std::string parName, TiXmlElement* e,
                              TString defaultValue = PARAMETER_NOT_FOUND_STR);
     Double_t GetDblParameterWithUnits(std::string parName, TiXmlElement* e,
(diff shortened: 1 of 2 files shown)


jgalan:
It looks as the compilation error is related with `Bool_t` and `bool` types? Perhaps the `const` is causing problems here?

```
/builds/rest-for-physics/framework/source/framework/tools/inc/TRestReflector.h:471:84: error: invalid conversion from 'Bool_t (*)(const string&)' {aka 'bool (*)(const std::__cxx11::basic_string<char>&)'} to 'bool (*)(std::string)' {aka 'bool (*)(std::__cxx11::basic_string<char>)'} [-fpermissive]
[388](https://gitlab.cern.ch/rest-for-physics/framework/-/jobs/23650145#L388)
```

source/framework/tools/inc/TRestStringHelper.h
```diff
@@ -36,7 +36,7 @@ Double_t StringToDouble(std::string in);
 Int_t StringToInteger(std::string in);
 std::string IntegerToString(Int_t n, std::string format = "%d");
 std::string DoubleToString(Double_t d, std::string format = "%4.2lf");
-Bool_t StringToBool(std::string in);
+Bool_t StringToBool(const std::string& in);
```
nkx111, line 39: The compilation failure is due to this `const` keyword. It makes the argument unmatched. If we think this change is necessary then we shall open a new PR to replace all the `AddConverter(xxx)` with const-input methods.

lobis: I think it should be implemented if it's not too much work, its not high priority but using `const T&` should always be used to pass non built-in types or atleast be the default option, so I think we should support it.

PR #278, opened 2022-07-31
Added fmt as external dependency and printing example

lobis:
lobis Medium: 107 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/lobis-external-deps) [](https://github.com/rest-for-physics/framework/commits/lobis-external-deps)

Added fmt (https://github.com/fmtlib/fmt) as dependency to help with string formatting.

I also moved `-lstdc++fs` outside of that block, I think it was a mistake (it should always be linked).

I added some simple class / methods in the `fmt` namespace and placed in `TRestTools.h/cpp` to help with printing.

Methods such as `fmt::ToEnergyString` take a number in the default REST energy scale and return the appropiate string, for example `fmt::ToEnergyString(10000)` returns `"10.00 MeV"`. Other methods such as `fmt::ToLengthString` have been added and its easy to implement more.

I updated the print for `TRestEvent` (and `TRestGeant4Event/Track` in another PR) to showcase what can be done with this library.

For instance one can define a custom formatter for any class, such as `TVector3` to simplify printing:
```
template <>
struct fmt::formatter<TVector3> : formatter<string> {
    auto format(TVector3 c, format_context& ctx) {
        string s = fmt::format("({:0.3f}, {:0.3f}, {:0.3f})", c.X(), c.Y(), c.Z());
        return formatter<string>::format(s, ctx);
    }
};

TVector3 myVector = {1,2,3};
fmt::print("{}", myVector); // will print "(1.000, 2.000, 3.000)"
```

Its very easy to add colors to the terminal which helps when reading the output:

image


nkx111:
Is this dependence necessary?

We already have `ToString()` method for `TVector3`, through `TRestReflector`, and colored output methods in `TRestStringOutput`. We can also modify this class for energy output methods. 

If fmt library is really better, I suggest to modify `TRestStringOutput` and try to adapt it. For example, we can make `RESTMetadata<<"xxx"<<RESTEndl;` calling `print(fg(color::green) | emphasis::bold, "xxx");` internally.

jgalan:
I think it is good that we keep a unified way to output things on screen. I.e. using `REST_xyz` macros, I think that is the point from @nkx111 in the last comment.

PR #289, opened 2022-09-01
Revert "fixed reading of REST v2.2 files"

jgalan:
jgalan Ok: 15 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/jgalan_fix_rawlib) [](https://github.com/rest-for-physics/framework/commits/jgalan_fix_rawlib)

This reverts commit 050d3bc29d904ca8807772739c700de6d3016af3.

This should fix rest-for-physics/rawlib#74

### DIFF
--- a/.github/workflows/validation.yml
+++ b/.github/workflows/validation.yml
@@ -76,29 +76,6 @@ jobs:
           path: ${{ env.REST_PATH }}
           retention-days: 1
 
-  filecompatibility:
-    name: "Previous Version File Compatibility"
-    runs-on: ubuntu-latest
-    container:
-      image: ghcr.io/lobis/root-geant4-garfield:rest-for-physics
-    needs: [ framework-install ]
-    steps:
-      - uses: actions/checkout@v3
-        with:
-          repository: rest-for-physics/framework
-          path: framework
-      - name: Restore cache
-        uses: actions/cache@v3
-        id: framework-install-cache
-        with:
-          key: ${{ env.BRANCH_NAME }}-${{ github.sha }}
-          path: ${{ env.REST_PATH }}
-      - name: Open V2.2 file
-        run: |
-          source ${{ env.REST_PATH }}/thisREST.sh
-          cd framework/pipeline/fileopening
-          restManager PrintTrees v2.2.30_hits.root
-
   metadata:
     name: "Metadata"
     runs-on: ubuntu-latest
--- a/.github/workflows/validation.yml
+++ b/.github/workflows/validation.yml
@@ -76,29 +76,29 @@ jobs:
           path: ${{ env.REST_PATH }}
           retention-days: 1
 
-  #  file-compatibility:
-  #    name: "Previous Version File Compatibility"
-  #    runs-on: ubuntu-latest
-  #    container:
-  #      image: ghcr.io/lobis/root-geant4-garfield:rest-for-physics
-  #    needs: [ framework-install ]
-  #    steps:
-  #      - uses: actions/checkout@v3
-  #        with:
-  #          repository: rest-for-physics/framework
-  #          path: framework
-  #      - name: Restore cache
-  #        uses: actions/cache@v3
-  #        id: framework-install-cache
-  #        with:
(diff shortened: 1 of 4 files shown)


source/framework/tools/inc/TRestReflector.h
```diff
@@ -242,13 +242,8 @@ inline TClass* GetClassQuick(std::string type) {
         return iter->second;
     } else {
         TClass* cl = TClass::GetClass(type.c_str());
-        if (cl != nullptr && cl->HasDictionary()) {
```
jgalan, line 245: @nkx Probably the `cl->HasDictionary()` here is a problem.

juanangp:
I think we should see this issue with a different perspective, we should understand why the dictionary is not needed while compiling `rawlib` with `detectorlib` dependencies.

The changes that you propose triggers the generation of the dictionary, I think that due to the changes that were done before this line was never reached.

I think that we should avoid the generation of a dictionary on the fly, so we should understand what is going on with the compilation. I think this issue should be fixed in the `cmake`.

jgalan:
I think the reason why the reverted commit introduces the problem, is because the type that has no dictionary is not added to the list, and therefore, it is not automatically compiled on the fly, since there is the condition `cl->HasDictionary()`.

I agree that could be a better way to do this, but personally I don't understand deeply all behind reflection.

Perhaps we could just have a list of supported types in REST somewhere, and generate the dictionaries at `cmake` as you propose.

juanangp:
> 

I think that the pipeline passes because we have a data member `std::map<int,double>` defined on `TRestDetectorSingleChannelAnalysisProcess.h`  in `detectorlib`, ideally we should generate an umbrella dictionary for all the possible observables that we have. Not sure how this can be implemented.

nkx111:
I think generating dictionaries on the fly is still the best way to solve this problem. Commit https://github.com/rest-for-physics/framework/commit/050d3bc29d904ca8807772739c700de6d3016af3 avoids problematic TClass objects being read from v2.2 files, but vetos the real-time generation functionality. I will try to fix it.

jgalan:
@nkx111 I don't understand why creating the required dictionaries at `cmake` stage would not be a solution.

Anyway, shall we merge this PR to fix pipelines at rawlib, then create an issue for further discussion?

nkx111:
Sorry for the delay. It's OK to revert the changes in TRestReflector. Now the file opening problem is also fixed from TRestRun side.

nkx111:
hi @jgalan, the pipelines are faling due to a certificate issue, can you check it out?

jgalan:
Hi, I forward the issue to the group IT. Anyhow, if we use the option `--no-check-certificate` for wget it will skip the warning and download, so I don't know if it is a problem or not to add this option @juanangp @lobis?

jgalan:
I believe now the pipeline is failing because `TFile::Open` also checks the certificate, I dont know how to bypass that, but the certificates will be updated soon.

jgalan:
Lets merge?

juanangp (approved):

lobis:
> I believe now the pipeline is failing because `TFile::Open` also checks the certificate, I dont know how to bypass that, but the certificates will be updated soon.

> Lets merge?

Since we cannot disable checking for certificates everywhere (`TFile`) I think we should keep it as it was, do you agree @rest-for-physics/core_dev ?

.github/workflows/validation.yml
```diff
@@ -253,7 +253,7 @@ jobs:
         run: |
           source ${{ env.REST_PATH }}/thisREST.sh
           cd framework/pipeline/trex
-          wget https://sultan.unizar.es/trexdm-readouts/readouts_v2.3.root
+          wget --no-check-certificate https://sultan.unizar.es/trexdm-readouts/readouts_v2.3.root
```
lobis, line 256:
```suggestion
          wget https://sultan.unizar.es/trexdm-readouts/readouts_v2.3.root
```

lobis: https://github.com/rest-for-physics/framework/pull/289#issuecomment-1236749856

lobis (approved):

PR #292, opened 2022-09-06
Adding loading macros validation

jgalan:
jgalan Ok: 49 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/jgalan_macros_validation) [](https://github.com/rest-for-physics/framework/commits/jgalan_macros_validation)

This PR adds a validation routine to check that macros are loaded without any "warning" or "error".


.github/workflows/validation.yml
```diff
@@ -99,6 +99,32 @@ jobs:
...
+        id: framework-install-cache
+        with:
+          key: ${{ env.BRANCH_NAME }}-${{ github.sha }}
+          path: ${{ env.REST_PATH }}
+      - name: Load restRootMacros
+        run: |
+          source ${{ env.REST_PATH }}/thisREST.sh
+          cd framework/pipeline
+          export DISPLAY=localhost:0.0
+          restRoot -b -q 2> error.log
```
juanangp, line 124:
```suggestion
          restRoot -b -q -m 1 2> error.log
```
I think you should do ´restRoot -m 1´ to load the macros, in fact now macros are giving warnings/errors

lobis:
Perhaps we should also add a way to check if all macros are loaded?

nkx111:
There are some errors from REST_Geant4_FindGammasEmitted.C and REST_Geant4_ViewEvent.C. Looks like some methods are renamed? @lobis

lobis:
> There are some errors from REST_Geant4_FindGammasEmitted.C and REST_Geant4_ViewEvent.C. Looks like some methods are renamed? @lobis 

Most likely, I will fix it.

jgalan:
Great! This validation test already hunted an issue before even being merged!

jgalan:
Finally pipeline is green, plz approve

.github/workflows/validation.yml
```diff
@@ -99,6 +99,31 @@ jobs:
...
+        uses: actions/cache@v3
+        id: framework-install-cache
+        with:
+          key: ${{ env.BRANCH_NAME }}-${{ github.sha }}
+          path: ${{ env.REST_PATH }}
+      - name: Load restRootMacros
+        run: |
+          source ${{ env.REST_PATH }}/thisREST.sh
+          cd framework/pipeline
+          export DISPLAY=localhost:0.0
```
juanangp, line 123: Why this line is needed?

jgalan: When there is no DISPLAY defined then I get a warning since I do not have proper X-forwarding configured. This generates a warning on my side, locally. I thought it could be a cause of troubles also at the docker. But I imagine it is not.

```
jgalan@sultan2:~/rest-framework/source/libraries$ restRootMacros 
Warning in <UnknownClass::SetDisplay>: DISPLAY not set, setting it to 192.168.1.2:0.0
```

juanangp: I don't think this is an issue in the docker but perhaps @lobis can comment on it.

This line seems weird to me, I rather prefer to remove it.

jgalan: There is no issue on the docker, and I don't think that the line seems weird to you is a solid argument to take out the line.

Setting that environment variable there is completely harmless, and it only has effect on that particular validation test. 

Don't see a real problem here, but. if you give me a good reason to take out the line, I will take it out.

lobis: I think adding the DISPLAY env variable is the easiest way to avoid the problem but I guess the most correct thing to do would be to load `restRoot` using batch mode or something similar, so that it doesn't care about displays.

I would think that the `-b` flag would solve this but apparently not so its probably not easy or possible to fix this warning, so I would keep it unless someone has a better idea.

jgalan: @lobis do you think it is possible to solve the Garfield output warnings? This might be related to a Garfield not compiled with the same ROOT version? I do not have those messages locally.

lobis: > @lobis do you think it is possible to solve the Garfield output warnings? This might be related to a Garfield not compiled with the same ROOT version? I do not have those messages locally.

There is only one root version installed on the image, and Garfield is installed after root (and with the `PATH`, `ROOTSYS` and `LD_LIBRARY_PATH` variables correctly referencing the root install. @juanangp and I tried to remove these warnings extensively for a while but didn't manage to do it.

If you want to try I think the easiest way would be to install Garfield and then REST on the official root docker image (`docker run --rm -it rootproject/root`) and see if you can fix it.

jgalan: Probably the problem is that we are not adding the Garfield headers to `ROOT_INCLUDE_PATH`

.github/workflows/validation.yml
```diff
@@ -99,6 +99,34 @@ jobs:
...
+          echo $GARFIELD_INSTALL
+          ls $GARFIELD_INSTALL
+          export ROOT_INCLUDE_PATH=$ROOT_INCLUDE_PATH:/usr/local/garfieldpp/install/Include
```
lobis, line 126: `ROOT_INCLUDE_PATH` already points to the Garfield install on the docker image, you can check this on the image itself looking at the environment variables.

lobis: For quick reference here are the env variables. I used `/bin/sh` as the shell to avoid loading variables from the .bashrc so I think this should be the exact same variables active when running the pipeline in a non-interactive shell:

```
# docker run -it --rm ghcr.io/lobis/root-geant4-garfield-dev:rest-for-physics /bin/sh
# printenv
GARFIELD_INSTALL=/usr/local/garfieldpp/install
G4INCLDATA=/usr/local/geant4/install/share/Geant4-11.0.3/data/G4INCL1.0
HOSTNAME=6d130f6981a5
LD_LIBRARY_PATH=/usr/local/geant4/install/lib:/usr/local/garfieldpp/install/lib:/usr/local/root/install/lib:
HOME=/root
G4PARTICLEXSDATA=/usr/local/geant4/install/share/Geant4-11.0.3/data/G4PARTICLEXS4.0
G4LEDATA=/usr/local/geant4/install/share/Geant4-11.0.3/data/G4EMLOW8.0
G4LEVELGAMMADATA=/usr/local/geant4/install/share/Geant4-11.0.3/data/PhotonEvaporation5.7
G4RADIOACTIVEDATA=/usr/local/geant4/install/share/Geant4-11.0.3/data/RadioactiveDecay5.6
G4PIIDATA=/usr/local/geant4/install/share/Geant4-11.0.3/data/G4PII1.3
G4SAIDXSDATA=/usr/local/geant4/install/share/Geant4-11.0.3/data/G4SAIDDATA2.0
G4REALSURFACEDATA=/usr/local/geant4/install/share/Geant4-11.0.3/data/RealSurface2.2
GARFIELD_VERSION=eb1b5403
GEANT4_VERSION=v11.0.3
TERM=xterm
ROOT_VERSION=v6-26-06
ROOTSYS=/usr/local/root/install
G4ABLADATA=/usr/local/geant4/install/share/Geant4-11.0.3/data/G4ABLA3.1
PATH=/usr/local/geant4/install/bin:/usr/local/root/install/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
HEED_DATABASE=/usr/local/garfieldpp/install/share/Heed/database
G4NEUTRONHPDATA=/usr/local/geant4/install/share/Geant4-11.0.3/data/G4NDL4.6
DEBIAN_FRONTEND=noninteractive
ROOT_INCLUDE_PATH=/usr/local/garfieldpp/install/include/Garfield:/usr/local/garfieldpp/install/include:
G4ENSDFSTATEDATA=/usr/local/geant4/install/share/Geant4-11.0.3/data/G4ENSDFSTATE2.3
CMAKE_CXX_STANDARD=17
PWD=/
PYTHONPATH=/usr/local/root/install/lib:
TINI_VERSION=v0.19.0
CMAKE_PREFIX_PATH=/usr/local/garfieldpp/install:
```

jgalan: However, when you `source thisREST.sh` it will be overwritten by the `ROOT_INCLUDE_PATH` definition found there.

jgalan: If that was the problem, the next commit should solve it.

jgalan: It must be `thisREST.sh` who sets that up, so that the user does not need to care about setting any env variables

cmake/thisREST.cmake
```diff
@@ -84,7 +84,7 @@ fi
 
 export REST_SOURCE=${CMAKE_CURRENT_SOURCE_DIR}
 export REST_PATH=\\\${thisdir}
-export ROOT_INCLUDE_PATH=\\\${thisdir}/include:${garfieldDir}/include
+export ROOT_INCLUDE_PATH=\\\$ROOT_INCLUDE_PATH:\\\${thisdir}/include
```
jgalan: No!!!

jgalan: You need to define the garfield path in ROOT_INCLUDE_PATH. It will only work in your docker

jgalan: For example, if you add `:${ROOT_INCLUDE_PATH}` behind as you did in next commits, you are hiding a problem in the validation pipeline, because you defined in your docker `ROOT_INCLUDE_PATH` externally, but this will not be the case for most of the users, so my opinion is that is better that we overwrite `ROOT_INCLUDE_PATH`.

Because the only scenario where this would cause problems is that someone else writes another ROOT library that has nothing to do with REST, and wants to use it. Ok, but I believe this will not be the case in the 99.9999%.

Somehow, we are conditioning the validation tests to the way you defined the variables inside the docker, which might not be the same for any user.

I agree with other env variables there is no other way, such as `LD_LIBRARY_PATH`

jgalan:
Should we approve and move to the next page?

lobis (approved):

PR #304, opened 2022-09-21
TRestPatternMask is not pure virtual anymore

jgalan:
jgalan Ok: 18 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/jgalan_new_masks) [](https://github.com/rest-for-physics/framework/commits/jgalan_new_masks)

`TRestPatternMask` defines common members for any `TRestPatternMask`. `TRestPatternMask` was pure virtual so in order to use one had to chose a particular pattern (Grid/Stripped/../Combined).

Now it is also possible to do not define a pattern, and thus there will be no pattern, the only selection will be the limits of the pattern defined by `fMaxRadius` and `fOffset`.

### DIFF
--- a/source/framework/core/inc/TRestEventProcess.h
+++ b/source/framework/core/inc/TRestEventProcess.h
@@ -213,6 +213,8 @@ class TRestEventProcess : public TRestMetadata {
 
     inline void SetObservableValidation(bool validate) { fValidateObservables = validate; }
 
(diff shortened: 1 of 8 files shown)


jgalan:
@nkx111  Any idea why the pipeline is failing here?

jovoy (approved):

juanangp (approved):

PR #323, opened 2022-11-10
Add direct support for REST classes in PyROOT (Python)

lobis:
lobis Ok: 72 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/lobis-python-bindings) [](https://github.com/rest-for-physics/framework/commits/lobis-python-bindings)

After this PR, REST classes can be imported into Python via

```python
import REST

# they can be accessed via (but not recommended)
run = REST.ROOT.TRestRun()

# this would be the preferred way to use REST classes from Python
import REST
import ROOT

run = ROOT.TRestRun()
```


juanangp (approved):

jgalan:
Thanks for this PR. I love it!

jgalan (approved):

jgalan:
When I do `import REST` in a python interactive session it takes long to load. I imagine this is because it is trying to load macros and not just libraries?

lobis:
> When I do `import REST` in a python interactive session it takes long to load. I imagine this is because it is trying to load macros and not just libraries?

It's because loading root takes a long time (`import ROOT`). `import REST` calls `import ROOT` behind the scenes (if it has not been called before, as other python imports) which takes a few seconds usually. If you do `import ROOT` and then `import REST` it should be instant I think.

This only loads REST libraries, you cannot use macros from python.

Vindaar:
What's the actual reason that the REST shared libraries need to be loaded only to _read_ a REST generated ROOT file anyway? Doesn't ROOT provide some interface to define custom types to store in a ROOT file (I assume that's what requires it)? 
I've once tried to be able to read from an `AnalysisTree` with pure ROOT and couldn't get anything sensible out of it (segfault land, thanks ROOT!). Or maybe there is some interface one can use to define the correct structs / whatever such that reading without REST is possible that I'm not aware of?

PR #326, opened 2022-11-14
CMakeLists. Adding USE_MPFR definition

jgalan:
A minor update required by rest-for-physics/axionlib#31

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -136,6 +136,7 @@ if (${REST_MPFR} MATCHES "ON")
     find_path(MPFR_INCLUDE mpfr.h HINTS ${MPFR_PATH} ${MPFR_PATH}/include) # path to include directory
     find_path(MPFR_LIB_PATH libmpfr.so HINTS ${MPFR_PATH} ${MPFR_PATH}/lib)
     find_library(MPFR_LIB mpfr HINTS ${MPFR_PATH} ${MPFR_PATH}/lib) # path to .so file
+	add_definitions(-DUSE_MPFR)
     if (NOT MPFR_INCLUDE MATCHES "MPFR_INCLUDE-NOTFOUND" AND NOT MPFR_LIB MATCHES "MPFR_LIB-NOTFOUND")
         # MPFR is found
         list(APPEND rest_features "MPFR")


jovoy (approved):

jgalan:
This PR just adds a definition that can be used to encapsulate code that requires MPFR libraries to be linked.

lobis (approved):
I updated to only have a single definition of `USE_MPFR`. The position of this definition is updated as a result of this PR, I have not verified this is necessary but I assume it is (I thought I had checked when I placed it under the `if` initially, but I don't remember the details).

If you think everything is working as expected it looks good to me @jgalan

CMakeLists.txt
```diff
@@ -136,6 +136,7 @@ if (${REST_MPFR} MATCHES "ON")
     find_path(MPFR_INCLUDE mpfr.h HINTS ${MPFR_PATH} ${MPFR_PATH}/include) # path to include directory
     find_path(MPFR_LIB_PATH libmpfr.so HINTS ${MPFR_PATH} ${MPFR_PATH}/lib)
     find_library(MPFR_LIB mpfr HINTS ${MPFR_PATH} ${MPFR_PATH}/lib) # path to .so file
+	add_definitions(-DUSE_MPFR)
```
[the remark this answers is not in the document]
lobis: For some reason this thread is not showing to all users...

CMakeLists.txt
```diff
@@ -136,10 +136,10 @@ if (${REST_MPFR} MATCHES "ON")
     find_path(MPFR_INCLUDE mpfr.h HINTS ${MPFR_PATH} ${MPFR_PATH}/include) # path to include directory
     find_path(MPFR_LIB_PATH libmpfr.so HINTS ${MPFR_PATH} ${MPFR_PATH}/lib)
     find_library(MPFR_LIB mpfr HINTS ${MPFR_PATH} ${MPFR_PATH}/lib) # path to .so file
+    add_compile_definitions(USE_MPFR)
```
jgalan, line 139: @lobis shouldn't be placed inside the second if condition?

jgalan: In my local copy it is inside the `if( NOT MPFR_INCLUDE ...` and it seems to be working properly

jgalan: If MPFR libraries are not found then code using MPFR libraries should not be exposed.

lobis: but isn't this the original location? that is why I changed it... you moved it in the first place no?

PR #334, opened 2022-11-24
ApplyCut will be always called

jgalan:
jgalan Ok: 4 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/jgalan_applyCut_always) [](https://github.com/rest-for-physics/framework/commits/jgalan_applyCut_always)

Not sure if this could be a performance issue, but I guess it will not be called in case `cutsEnabled` is `false`.


jgalan:
I introduced `ApplyCut` at each loop where `ProcessEvent` is called. I was not sure if this was necessary at any of them. @nkx

lobis (approved):

source/framework/core/src/TRestThread.cxx
```diff
@@ -163,6 +163,7 @@ bool TRestThread::TestRun() {
 
             fProcessChain[j]->BeginOfEventProcess(ProcessedEvent);
             ProcessedEvent = fProcessChain[j]->ProcessEvent(ProcessedEvent);
+            if (fProcessChain[j]->ApplyCut()) ProcessedEvent = nullptr;
```
juanangp, line 166: Don't quite understand the logic here, I think we should review this class

Vindaar: It simply calls the `ApplyCut` method of each process (and each event). If the event is _removed_ by the cut, it is not kept, hence a `nullptr` is returned.

juanangp: I do understand the what it is suppose to do, what I don't understand is the following:
 - The need of the `if` `else` statement for debugging here https://github.com/rest-for-physics/framework/blob/4cf0a2e9a0a6f81f85226159f964a7040da05159/source/framework/core/src/TRestThread.cxx#L488
 - The use of `size` instead of `std` iterators https://github.com/rest-for-physics/framework/blob/4cf0a2e9a0a6f81f85226159f964a7040da05159/source/framework/core/src/TRestThread.cxx#L547
 - The user case of `UseTestRun` https://github.com/rest-for-physics/framework/blob/4cf0a2e9a0a6f81f85226159f964a7040da05159/source/framework/core/src/TRestThread.cxx#L557
 - The need of `CloneTo`, which seems rather complicated, might be replaced by `std::copy` or `std::move` https://github.com/rest-for-physics/framework/blob/4cf0a2e9a0a6f81f85226159f964a7040da05159/source/framework/core/src/TRestEvent.cxx#L96
 
My point is that I think that we should review this class, I am not against this particular PR.

jgalan: Yes, I agree also about revision of this class.

This PR simply forces a call to `ApplyCut` after each `ProcessEvent` instead of having to implement it at each `ProcessEvent`.

The topics brought by @juanangp should be discussed in a dedicated issue. I hope @nkx could also say something here, mostly about the TestRun. I believe this is connected about the custom observables names that are defined at some processes, but myself never went too deep on this matter.

jgalan: I will pay 5 beers on Xmass dinner to the first one solving this issue #338

jgalan:
Perhaps we should merge as soon as @cmargalejo states the cuts are working on her side

PR #342, opened 2022-11-30
Additional institution logos added to README.md

jgalan:


nkx111:
It's OK to put PandaX-III logo in README. We use same logo as PandaX collaboration, which is:
logo

nkx111 (approved):

lobis (approved):

PR #343, opened 2022-12-07
TRestAnalysisTree. Reviewing documentation

jgalan:
jgalan Ok: 36 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/jgalan_tree_documentation) [](https://github.com/rest-for-physics/framework/commits/jgalan_tree_documentation)

The TRestAnalysisTree, together with the TRestMetadata class are one of the most visited pages on our API documentation pages. Thus, I took a bit of time to review.

Still, it might require your attention to improve the description of the class.


source/framework/core/src/TRestAnalysisTree.cxx
```diff
@@ -22,36 +22,44 @@
...
 //////////////////////////////////////////////////////////////////////////
 ///
-/// TRestAnalysisTree is TTree but with **managed objects** for the branches to fill.
-/// There are six fixed branches of event information in TRestAnalysisTree: runOrigin,
-/// subRunOrigin, eventID, subEventID, subEventTag and timeStamp. They are pointing
-/// to the corresponding class members inside TRestAnalysisTree. Those branches are
-/// called `event branches`. Additional branches could be added by the user, they will
-/// point to some objects whose addresses are also stored in this class. Those objects
-/// are called `observables`.
+/// TRestAnalysisTree is a TTree but with **managed objects** for the branches to be
```
jgalan, line 25: What do we understand by `managed` objects here? There is a better word?

source/framework/core/src/TRestAnalysisTree.cxx
```diff
@@ -22,36 +22,44 @@
...
+/// `TTree::Fill()` to create and save a new entry inside the tree.
 ///
-/// In TRestAnalysisTree, the concept of "Branch" is weakened. We can directly call
-/// `SetObservableValue()` and then `TRestAnalysisTree::Fill()` to do the data saving.
-/// The code could be simplified while sacrificing a little performance. We can use
+/// In TRestAnalysisTree, the concept of "Branch" is weakened. We update the
+/// variables by invoking the `SetObservableValue()` method and then
+/// TRestAnalysisTree::Fill() to generate a new entry inside the tree.
+/// The code will be simplified while sacrificing a little performance. We can use
 /// temporary variable to set observable value directly. We can focus on the analysis
```
jgalan, line 43: I do not catch the meaning of these 2 last sentences.

In one hand what we do is to encapsulate the use of the tree through the method `SetObservableValue` so that the user does not need to be playing around with variable generation and branch address creation. Perhaps is something like that we should say here?

nkx111: > what we do is to encapsulate the use of the tree through the method SetObservableValue so that the user does not need to be playing around with direct variable generation and branch address creation

Exactly. That shall be one of the selling point of `TRestAnalysisTree`. We can clarify it in the text.

nkx111 (approved):

source/framework/core/src/TRestAnalysisTree.cxx
```diff
@@ -22,36 +22,44 @@
...
 ///
-/// TRestAnalysisTree is TTree but with **managed objects** for the branches to fill.
-/// There are six fixed branches of event information in TRestAnalysisTree: runOrigin,
-/// subRunOrigin, eventID, subEventID, subEventTag and timeStamp. They are pointing
-/// to the corresponding class members inside TRestAnalysisTree. Those branches are
-/// called `event branches`. Additional branches could be added by the user, they will
-/// point to some objects whose addresses are also stored in this class. Those objects
-/// are called `observables`.
+/// TRestAnalysisTree is a TTree but with **managed objects** for the branches to be
+/// filled. Inside a TRestAnalysisTree we find always the following six branches
```
jgalan, line 26:
```suggestion
/// filled. The user will decide ins each event data processing chain which branches/observables/variables
will be finally added to the analysis tree. Inside a TRestAnalysisTree we find always the following six branches
```

jgalan:
Please, approve!

PR #345, opened 2022-12-08
Implementation of TRestDataSet

jgalan:
jgalan Large: 819 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/jgalan_dataset) [](https://github.com/rest-for-physics/framework/commits/jgalan_dataset)

This PR is related to issue #13.

A class allowing to create datasets from an official REST data repository.

The class documentation has been pre-rendered at the following location: https://sultan.unizar.es/test/classTRestDataSet.html have a look!

RML setup would look as follows:

```
    <TRestDataSet name="BabyIAXO_Test_DataSet_00">
        <parameter name="startTime" value="2022/04/28 00:00" />
        <parameter name="endTime" value="2022/11/25 13:20" />
        <parameter name="filePattern" value="/home/jgalan/rest-framework/source/libraries/axion/examples/full-ray-tracing/BabyIAXO_TrueWolterMicromegasTest*.root"/>

        <filter metadata="TRestRun::fRunNumber" greaterThan="370" lowerThan="375" />
        <filter metadata="TRestRun::fRunTag" contains="Test" />

        <observables list="window_transmission:optics_efficiency:axionPhoton_probability:axionPhoton_transmission:boreExitGate_transmission" />

        <processObservables list="final" />
    </TRestDataSet>
```

The class will be capable to generate a final txt file with the selected data, and a header with the origin of the data:

```
### TRestDataSet generated file
###
### StartTime : 2022/04/28 00:00
### EndTime : 2022/11/25 13:20
###
### Data path : /home/jgalan/rest-framework/source/libraries/axion/examples/full-ray-tracing
### File pattern : BabyIAXO_TrueWolterMicromegasTest*.root
###
### Metadata filters :
### - TRestRun::fRunNumber. Greater than: 370.000. Lower than: 375.000.
### - TRestRun::fRunTag. Contains: Test.
###
### Observables list:  runOrigin eventID timeStamp window_transmission optics_efficiency axionPhoton_probability axionPhoton_transmission boreExitGate_transmission final_energy final_posX final_posY final_posZ final_R final_thetaAngle final_phiAngle
###
### Data starts here
000373  000003  0000000000  9.945e-01   2.828e-01   4.034e-19   1.000e+00   1.000e+00   7.021e+00   5.524e+00   -5.612e-01  1.454e+04   5.553e+00   4.188e-02   1.313e+00
000373  000004  0000000000  0.000e+00   6.521e-01   7.624e-19   1.000e+00   1.000e+00   2.744e+00   2.122e+00   -1.710e+00  1.454e+04   2.726e+00   2.012e-02   2.235e+00
```


source/framework/core/src/TRestDataSet.cxx
```diff
@@ -321,6 +322,7 @@ std::vector<std::string> TRestDataSet::FileSelection() {
 
         if (!accept) continue;
 
+        fTotalDuration += run->GetEndTimestamp() - run->GetStartTimestamp();
```
cmargalejo, line 325: Maybe it would be good to be able to define the run duration but with a minimum number of events per minute (e.g. 5 or 10 events per minute), so that we don't take into account time in which for example there was a trip and data was not being taken. This way we would not compute the run duration but the effective time of data taking during the run.

cmargalejo (approved):

source/framework/core/src/TRestDataSet.cxx
jgalan: Right, however, here we are either accepting the file or rejecting the file. This variable registers the total duration from files added to the list.

You could reject the entire file by adding a metadata rule such that the mean rate, captured for example as a metadata via `TRestSummaryProcess`, would exceed certain amount.

However, if rejecting one file is not a solution because the file is the full tracking (for example) and what you want is to partially accept the events inside a given file, then you need to do that later on.

This is because of the nature of adding files inside RDataFrame.

This might need a dedicated method running on the final tree where you find those gaps of data, and reduce the `fTotalDuration` accordingly.

Adding something such as `TRestDataSet::RemoveDataUnderConditions( std::string )`. Probably you need to loop.

I added inside the issue.

KonradAltenmueller (approved):

source/framework/core/inc/TRestDataSet.h
```diff
@@ -0,0 +1,103 @@
...
+    /// A list populated by the FileSelection method using the conditions of the dataset
+    std::vector<std::string> fFileSelection;  //!
+
+    void InitFromConfigFile() override;
+
+   protected:
+    virtual std::vector<std::string> FileSelection();
+
+   public:
+    ROOT::RDataFrame GetDataFrame() {
```
lobis, line 78: I am not sure but this may be problematic, we are returning a new `ROOT::RDataFrame` object, instead of a reference to the dataset (which I assume is the intention).

Maybe what we want is

```suggestion
    ROOT::RDataFrame& GetDataFrame() {
```

instead?

It may not be a problem at the end because I think `RDataFrame` does not own the data, but I am not sure.

lobis:
Maybe this can close https://github.com/rest-for-physics/framework/issues/13?

source/framework/core/inc/TRestDataSet.h
```diff
@@ -85,6 +88,8 @@ class TRestDataSet : public TRestMetadata {
     Int_t GetNumberOfColumns() { return fDataSet.GetColumnNames().size(); }
     Int_t GetNumberOfBranches() { return GetNumberOfColumns(); }
 
+    Double_t GetTotalTimeInSeconds() { return fTotalDuration; }
```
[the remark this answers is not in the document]
jgalan: Yes, I replied there, I think that it would be easy once the dataset has loaded to integrate those periods time, subtract the time from total time, and remove those events using a standard TCut or Filter to replace the existing DataFrame/TTree. 

I added this issue to #346 in case you wish to take over

source/framework/core/inc/TRestDataSet.h
jgalan: I think we can just accept the new version with the reference. I don't think it makes a copy, clearly.

For me it is ok, but why I cannot press the button commit suggestion?

source/framework/core/src/TRestDataSet.cxx
```diff
@@ -0,0 +1,543 @@
...
+        return;
+    }
+
+    if (fFileSelection.empty()) FileSelection();
+
+    // We are not ready yet
+    if (fFileSelection.empty()) return;
+
+    ///// Disentangling process observables --> producing finalList
+    TRestRun* run = new TRestRun(fFileSelection[0]);
```
juanangp, line 227: I think there is no reason to use a pointer here, we should avoid use `new` and `delete`

source/framework/core/src/TRestDataSet.cxx
```diff
@@ -0,0 +1,543 @@
...
+    std::vector<std::string> fileNames = TRestTools::GetFilesMatchingPattern(fFilePattern);
+
+    if (!fileNames.empty()) {
+        RESTInfo << " - Starting to select files matching pattern. Total files : " << fileNames.size()
+                 << RESTendl;
+    }
+
+    fTotalDuration = 0;
+    for (const auto& file : fileNames) {
+        TRestRun* run = new TRestRun(file);
```
juanangp, line 296: Again `new` and `delete`, please use `TRestRun run(file);`

jgalan: For me when I do `new` and `delete` it is clear the instance is created and destroyed. Don't see why this is a problem here.

jgalan: I just want to make sure the instance is destroyed.

jgalan:
> I think it's important that unit tests are added for this class, now that we have the possibility to do it from the ground up.

I agree, just waiting to see what happens with the files I prepared for testing the class.

See https://root-forum.cern.ch/t/checkbytecount-and-streamer-error/52716/7

source/framework/core/src/TRestDataSet.cxx
juanangp: We should minimize the use of `new` and there is no reason to use it here
https://stackoverflow.com/questions/6500313/why-should-c-programmers-minimize-use-of-new

source/framework/core/inc/TRestDataSet.h
juanangp: I think you should avoid to return a reference you can either return a pointer or a copy, note that to return a copy is generally optimized.
https://stackoverflow.com/questions/2939165/c-get-method-returning-by-value-or-by-reference

examples/dataset.rml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0"?>
+<TRestDataSet name="BabyIAXO">
+    <parameter name="startTime" value="2022/04/28 00:00"/>
+    <parameter name="endTime" value="2022/11/25 13:20"/>
+    <parameter name="filePattern" value="../../framework-data/pipeline/dataset/BabyIAXO_TrueWolterMicromegasTest*.root"/>
+    <filter metadata="TRestRun::fRunNumber" greaterThan="370" lowerThan="375"/>
+    <filter metadata="TRestRun::fRunTag" contains="Test"/>
+    <observables list="window_transmission:optics_efficiency:axionPhoton_probability:axionPhoton_transmission:boreExitGate_transmission"/>
```
nkx111, line 8: why use `:` as the separator? If it's correlated with a vector we can use rml auto parsing, like: `{window_transmission,optics_efficiency,axionPhoton_probability,axionPhoton_transmission,boreExitGate_transmission}`

jgalan: Right, `,` separator is probably much better. I thought about it, perhaps we could use a `Split` automatic separator identifier?

Such as we find on string, if we find only `,`, we use it.

We could use a wildcard Split separator to try to find the separator.

jgalan: Just updated it at [b388349](https://github.com/rest-for-physics/framework/pull/345/commits/b3883496d98e06af576ffb5cbc061da62f430b14)

PR #350, opened 2022-12-13
Cmake fixes to clean up warnings during compilation

jgalan:
jgalan Large: 1441 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/jgalan_cmake_fix) [](https://github.com/rest-for-physics/framework/commits/jgalan_cmake_fix)

This PR will avoid some warnings that appear during compilation. 

Some warnings seem to be related to the fact that we were generating dictionaries for classes that do not inherit from Object. This is my guess.

- Updated `MacroRootDict.cmake` so that dictionary is not generated for some classes that do not inherit from TObject, and/or just do not need dictionary.

I implemented a patch the best I could at `make/MacroRootDict.cmake`, do not hesitate to improve the implementation of the present patch if you find a better way.

Now **compilation warnings will be considered errors so the pipeline will fail in case there are any warnings**.

IMPORTANT. There are still a number of warnings that are hidden from the user, because I didn't  manage to solve.

We may compile using `-DCMAKE_BUILD_TYPE=Debug` to show all warnings.

Other related PRs that MUST be merged together with this PR.

rest-for-physics/restG4#89
rest-for-physics/geant4lib#82
rest-for-physics/axionlib#43
rest-for-physics/tracklib#26
rest-for-physics/wimplib#4
rest-for-physics/rawlib#90
rest-for-physics/detectorlib#64
rest-for-physics/connectorslib#22


CMakeLists.txt
```diff
@@ -16,11 +16,20 @@ elseif (CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
 endif ()
 
 if (CMAKE_SYSTEM_NAME MATCHES "Darwin")
-    SET(CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS} "-fsanitize=address")
+    SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=address")
 else ()
-    SET(CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS} "-lstdc++fs")
+	SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -lstdc++fs")
```
lobis, line 21:
```suggestion
    SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -lstdc++fs")
```

source/framework/analysis/inc/TRestDataQualityRules.h
```diff
@@ -69,28 +69,28 @@ class TRestDataQualityRules : public TObject {
...
             return fValues[n];
     }
 
-    TVector2 GetRange(Int_t n) {
+    TVector2 GetRange(unsigned int n) {
         if (fRanges.size() <= n)
             return TVector2(0, 0);
         else
             return fRanges[n];
     }
```
lobis, line 91: I guess all these methods could be rewritten in a more compact way such as:

```suggestion
    TVector2 GetRange(unsigned int n) {
        return (n < fRanges.size()) ? fRanges[n] : {0, 0};
    }
```

source/framework/analysis/src/TRestRealTimeDrawingProcess.cxx
```diff
@@ -254,7 +254,7 @@ void TRestRealTimeDrawingProcess::DrawWithNotification() {
                 int _runNumber = run->GetRunNumber();
                 delete run;
                 if (_runNumber == runNumber) {
-                    for (int i = 0; i < fPlots.size(); i++) {
+                    for (unsigned int i = 0; i < fPlots.size(); i++) {
                         fPlots[i]->SetFile(message);
                         fPlots[i]->PlotCombinedCanvas();
                     }
```
lobis, line 260: In general we should try to refactor similar statements whenever we find the time (just a reminder).

```suggestion
                    for (auto& plot: fPlots) {
                        plot->SetFile(message);
                        plot->PlotCombinedCanvas();
                    }
```

source/framework/external/tinyxml/tinyxmlparser.cpp
```diff
@@ -41,11 +41,7 @@ distribution.
 // is less flexible than it appears. Changing the entries
 // or order will break putstring.
 TiXmlBase::Entity TiXmlBase::entity[TiXmlBase::NUM_ENTITY] = {
-    {"&amp;", 5, '&'},
-    {"&lt;", 4, '<'},
-    {"&gt;", 4, '>'},
-    {"&quot;", 6, '\"'},
-    {"&apos;", 6, '\''}};
+    {"&amp;", 5, '&'}, {"&lt;", 4, '<'}, {"&gt;", 4, '>'}, {"&quot;", 6, '\"'}, {"&apos;", 6, '\''}};
```
juanangp, line 44: I think we shouldn´t perform clang formatting in external classes.

source/framework/sensitivity/inc/TRestLikelihood.h
```diff
@@ -0,0 +1,71 @@
...
+ * For the list of contributors see $REST_PATH/CREDITS.                  *
+ *************************************************************************/
+
+#ifndef REST_TRestLikelihood
+#define REST_TRestLikelihood
+
+#include "TRestMetadata.h"
+
+/// UPDATE Write here a brief description. Just one line!
+class TRestLikelihood : public TRestMetadata {
```
juanangp, line 29: Thi is a new class? I think shouldn't be added under this PR

source/framework/tools/src/TRestStringHelper.cxx
```diff
@@ -132,7 +132,7 @@ string REST_StringHelper::ReplaceMathematicalExpressions(string buffer, Int_t pr
         erased = true;
     }
 
-    for (int i = 0; i < Expressions.size(); i++) {
+    for (unsigned int i = 0; i < Expressions.size(); i++) {
```
lobis, line 135: We should try to use `unsigned int` or `size_t` more consistently. My suggestion is to use `size_t` whenever we are dealing with the length of an array such as in this case and use `unsigned int` on the remaining cases.

```suggestion
    for (size_t i = 0; i < Expressions.size(); i++) {
```

source/framework/tools/src/TRestStringHelper.cxx
```diff
@@ -506,27 +506,17 @@ string REST_StringHelper::ToDateTimeString(time_t time) {
...
-        secondStr[2] = '\0';
-        secondStr[1] = secondStr[0];
-        secondStr[0] = '0';
-    }
-    char s[20];
-    sprintf(s, "%s-%s-%s %s:%s:%s", yearStr, monthStr, dayStr, hourStr, minuteStr, secondStr);
-    string str(s);
-    return str;
+
+    std::string yearStr = IntegerToString(year);
```
lobis, line 510: I think we should avoid explicitly calling the namespace `std::` in the source files (use `using namespace std;` instead), especially when using well-known types such as `string`. At least to remain coherent with the other code.

```suggestion
    string yearStr = IntegerToString(year);
```

source/framework/tools/src/TRestTools.cxx
```diff
@@ -1070,7 +1068,11 @@ int TRestTools::DownloadRemoteFile(string remoteFile, string localFile) {
     if ((string)url.GetProtocol() == "https" || (string)url.GetProtocol() == "http") {
         string path = TRestTools::SeparatePathAndName(localFiletmp).first;
         if (!TRestTools::fileExists(path)) {
-            system(("mkdir -p " + path).c_str());
+            int r = system(("mkdir -p " + path).c_str());
+            if (r != 0) {
+                std::cout << "mkdir failed to create directory: " << path << std::endl;
+                return -1;
+            }
```
lobis, line 1075: Since we are using C++17 we should use the `<filesystem>` API whenever we need to do interface with the filesystem such as in this case. (may need to `#include <filesystem>`).

```suggestion
            if (!filesystem::create_directories(path)){
                std::cerr << "mkdir failed to create directory: " << path << std::endl;
                return -1;
            }
```

lobis: Also I think we should use `std::cerr` to use the standard error stream instead of the standard output whenever there is an error.

source/framework/sensitivity/inc/TRestLikelihood.h
jgalan: Right, this ended here by error! There is a new draft PR dedicated to that here https://github.com/rest-for-physics/framework/pull/355

source/framework/external/tinyxml/tinyxmlparser.cpp
jgalan: Removed here 6a3055627424c8cf01eca47ca3144f7a50425650

source/framework/sensitivity/inc/TRestLikelihood.h
jgalan: Removed here 6a3055627424c8cf01eca47ca3144f7a50425650

source/framework/external/tinyxml/tinyxmlparser.cpp
jgalan: why?

jgalan: This is external but I believe modified by us, by @nkx

source/framework/analysis/inc/TRestDataQualityRules.h
```diff
@@ -69,28 +69,25 @@ class TRestDataQualityRules : public TObject {
...
             return fValues[n];
     }
 
-    TVector2 GetRange(Int_t n) {
-        if (fRanges.size() <= n)
-            return TVector2(0, 0);
-        else
-            return fRanges[n];
+    TVector2 GetRange(unsigned int n) {
+        return (n < fRanges.size()) ? fRanges[n] : {0, 0};
```
jgalan, line 87:
```suggestion
        return (n < fRanges.size()) ? fRanges[n] : TVector2(0,0);
```

cmargalejo (approved):

juanangp (approved):

PR #360, opened 2023-01-09
Add pre-commit hooks

lobis:
lobis Large: 144677 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/lobis-precommit) [](https://github.com/rest-for-physics/framework/commits/lobis-precommit)

The goal of this PR is to integrate with [pre-commit](https://pre-commit.com/), which in my experience is the current standard for managing pre-commit hooks.

pre-commit can be installed via pip (see documentation at the link above) and it has a very nice ecosystem of multi-language hooks which are really useful: clang-format, cmake-format, python formatting (black), etc. It also allows to define custom hooks locally which may be useful.

After installation the hooks will run after a commit only on the stage files.

I have also added a step in the CI to run `pre-commit` resulting in a pipeline failure if things such as the formatting of the code is not correct because a user failed to run clang-format manually or didn't use the hooks (because they were not installed or some other reason).

In the future I will also add the integration with pre-commit bot (not sure how to do this yet or how much work it could take). This will result in automatic formatting (with the bot as author) of code whenever a user fails to format the code properly in a PR.

Besides formatting, we could also add linting to the pre-commit, however this will require a very tedious revision of the code in order to pass since we haven't run any linters up to this point on the whole codebase.

In this PR I have also modified the relevant files in order to pass the formatting (python and cmake at the time of writting). These are only formatting changes. The style of formatting can be altered with configuration files.


jgalan:
I thought the code was already formatted automatically, so no manual formatting was required

lobis:
> I thought the code was already formatted automatically, so no manual formatting was required

We implemented a very rudimentary version of this utility via cmake (copy files into the hooks directory) but this relies on the user installing the required packages on its own (clang-format, xmlint, etc.). If we knew this existed we would never have done that.

pre-commit is much more powerful and easier, it uses python virtual environments so the user does not have to install anything (only pre-commit) and it allows for more powerful hooks such as and xml formatting tool that actually takes options or cmake formatting. Also the current implementation of hooks does not work for remote development and gives a cmake warning when not working.

I would remove the previous implementation in favour of this one.

jgalan:
@lobis what it is left for this to go through?

juanangp (approved):

jgalan (approved):

jgalan:
Large: 144677 lines!

jgalan:
@lobis plz, could you add some instructions in a new page "Pushing to the repository" at https://rest-for-physics.github.io/rest-advanced with what we need to do in our repositories, the two lines we need to execute basically.

PR #361, opened 2023-01-12
Upgrading root installation to root-6.26.10

juanangp:
juanangp Ok: 36 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/root-6.26.10) [](https://github.com/rest-for-physics/framework/commits/root-6.26.10)

- [x] Upgrading installation from root-6.24.2 to root-6.26.10, minor changes added on cmake flags
- [x] `scripts/installation/installGarfield.sh` updated to tags/4.0. Also reviewed and removed GARFIELD_HOME which is obsolete now.

### DIFF
--- a/scripts/installation/installROOT.sh
+++ b/scripts/installation/installROOT.sh
@@ -12,7 +12,7 @@ case "${unameOut}" in
     *)          machine="UNKNOWN:${unameOut}"
 esac
 
-ROOT_VERSION=6.24.02
+ROOT_VERSION=6.26.10
 ROOT_DIR=$HOME/apps/root-$ROOT_VERSION
 
 mkdir -p ${ROOT_DIR}
@@ -34,12 +34,11 @@ tar -xvzf $f
 rm -rf source
 mv root-$ROOT_VERSION source
 
-mkdir -p ${ROOT_DIR}/build
-cd ${ROOT_DIR}/build
+mkdir -p ${ROOT_DIR}/root_build
+cd ${ROOT_DIR}/root_build
 
-cmake -Wno-dev -Dbuiltin_glew=ON -DCMAKE_CXX_STANDARD=17 -Dgdml=ON -DCMAKE_INSTALL_PREFIX=${ROOT_DIR}/install  ../source
-make -j30
-make install
+cmake -DCMAKE_CXX_STANDARD=17 -DCMAKE_INSTALL_PREFIX=${ROOT_DIR}/install  ../source
+make -j8 install
 
 cd ${CURRENT_DIR}
 
--- a/scripts/installation/installROOT.sh
+++ b/scripts/installation/installROOT.sh
@@ -37,7 +37,7 @@ mv root-$ROOT_VERSION source
 mkdir -p ${ROOT_DIR}/root_build
 cd ${ROOT_DIR}/root_build
 
-cmake -DCMAKE_CXX_STANDARD=17 -DCMAKE_INSTALL_PREFIX=${ROOT_DIR}/install  ../source
+cmake -Wno-dev -Dbuiltin_glew=ON -DCMAKE_CXX_STANDARD=17 -Dgdml=ON -DCMAKE_INSTALL_PREFIX=${ROOT_DIR}/install  ../source
 make -j8 install
 
 cd ${CURRENT_DIR}
--- a/scripts/installation/installROOT.sh
+++ b/scripts/installation/installROOT.sh
@@ -35,7 +35,7 @@ rm -rf source
 mv root-$ROOT_VERSION source
 
 mkdir -p ${ROOT_DIR}/root_build
-cd ${ROOT_DIR}/root_build
+cd ${ROOT_DIR}/build
 
 cmake -Wno-dev -Dbuiltin_glew=ON -DCMAKE_CXX_STANDARD=17 -Dgdml=ON -DCMAKE_INSTALL_PREFIX=${ROOT_DIR}/install  ../source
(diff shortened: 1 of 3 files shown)


juanangp:
I have tested on Ubuntu 20.04 (WSL) and it is working properly @jgalan  perhaps you can test on MAC to check that it is working as well?

scripts/installation/installROOT.sh
```diff
@@ -34,12 +34,11 @@ tar -xvzf $f
...
-mkdir -p ${ROOT_DIR}/build
-cd ${ROOT_DIR}/build
+mkdir -p ${ROOT_DIR}/root_build
+cd ${ROOT_DIR}/root_build
 
-cmake -Wno-dev -Dbuiltin_glew=ON -DCMAKE_CXX_STANDARD=17 -Dgdml=ON -DCMAKE_INSTALL_PREFIX=${ROOT_DIR}/install  ../source
-make -j30
-make install
+cmake -DCMAKE_CXX_STANDARD=17 -DCMAKE_INSTALL_PREFIX=${ROOT_DIR}/install  ../source
+make -j8 install
```
lobis, line 41:
```suggestion
make -j install
```

Perhaps we can just us the `-j` option that uses the max. number of available threads, this way it can be independant of the system.

juanangp: While doing that my computer got stuck since it is going out of resources, perhaps we can check `nproc` and remove some cores?

jgalan: without a number it might take all the CPUs

jgalan: We might use: `cat /proc/cpuinfo | grep processor | wc -l`? And take half the number of CPUs?

jgalan:
> I have tested on Ubuntu 20.04 (WSL) and it is working properly @jgalan perhaps you can test on MAC to check that it is working as well?

I go for it

lobis:
> I have tested on Ubuntu 20.04 (WSL) and it is working properly @jgalan perhaps you can test on MAC to check that it is working as well?

I will test REST compilation with root binaries (6.22.10) on Ubuntu 22.04 (WSL). I think this should be the recommended installation method for root if it's not already (if it works).

scripts/installation/installROOT.sh
lobis: I updated my suggestion to use all minus 1 cores, I remember on some other project's documentation this is the recommended way, I have never seen half the number of cpu as recommended anywhere but we can also try it if this doesn't work.

juanangp: I have 12 cores in my WSL, but I already had issues using 10, so I think the safest is to go to the half of the cores.

juanangp:
> Ubuntu 22.04 (WSL)

I am targeting Ubuntu 20.04 and root 6.26.10. I was not able to find Ubuntu 22.04 (WSL) using command prompt, but I see it is available in Microsoft Store.

jgalan:
For info I got the following error on MacOs

```
-- Found GLEW: /usr/local/lib/cmake/glew/glew-config.cmake  
CMake Error at cmake/modules/SearchInstalledSoftware.cmake:614 (message):
  Please enable builtin Glew due bug in latest CMake (use cmake option
  -Dbuiltin_glew=ON).
Call Stack (most recent call first):
  CMakeLists.txt:245 (include)
```

juanangp:
> For info I got the following error on MacOs
> 
> ```
> -- Found GLEW: /usr/local/lib/cmake/glew/glew-config.cmake  
> CMake Error at cmake/modules/SearchInstalledSoftware.cmake:614 (message):
>   Please enable builtin Glew due bug in latest CMake (use cmake option
>   -Dbuiltin_glew=ON).
> Call Stack (most recent call first):
>   CMakeLists.txt:245 (include)
> ```

Ok, I just revert the changes on cmake params, it is weird since it is working properly on Linux

jgalan:
Ok, for me in MacOs, after adding ` -Dbuiltin_glew=ON` to the cmake works. It must be noted that `glew-dev` libraries were installed on my system.

jgalan (approved):
Works for me on MacOs Ventura 13.1

jgalan:
I also checked `./installGarfield.sh` and worked properly on MacOs

scripts/installation/installROOT.sh
```diff
@@ -34,12 +34,11 @@ tar -xvzf $f
 rm -rf source
 mv root-$ROOT_VERSION source
 
-mkdir -p ${ROOT_DIR}/build
-cd ${ROOT_DIR}/build
+mkdir -p ${ROOT_DIR}/root_build
+cd ${ROOT_DIR}/root_build
```
jgalan, line 38: Why we need to change to `root_build`? Inside the `root-6.26.10` directory there will be 3 directories, `build` `install` and `source`.

scripts/installation/installROOT.sh
jgalan: SO, cannot we do `($(nproc)/2)`

scripts/installation/installROOT.sh
jgalan, line 38:
```suggestion
cd ${ROOT_DIR}/build
```

scripts/installation/installROOT.sh
jgalan: So, `make -j8` is probably ok, perhaps even `make -j4`. Better to avoid problems.

lobis (approved):

PR #362, opened 2023-01-12
use [[(]]) instead of <<(>>) for analysis plots to correct invalid xml syntax

lobis:
lobis Ok: 75 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/lobis-xml) [](https://github.com/rest-for-physics/framework/commits/lobis-xml)

As the title says we cannot use `<<` or `>>` in the xml (or rml) files if we want it to be valid xml, even though some parsers can tolerate it.

I replaced them by square brackets but any suggestion is welcome.

This issue was discovered when running a tool to check for valid xml files.

### DIFF
--- a/source/framework/core/src/TRestAnalysisPlot.cxx
+++ b/source/framework/core/src/TRestAnalysisPlot.cxx
@@ -757,7 +757,7 @@ void TRestAnalysisPlot::PlotCombinedCanvas() {
                 RESTInfo << "AnalysisTree->Draw(\"" << plotString << ">>" << histoName << "\", \""
                          << cutString << "\", \"" << optString << "\", " << fDrawNEntries << ", "
                          << fDrawFirstEntry << ")" << RESTendl;
-                outVal = tree->Draw(plotString + "]]" + histoName, cutString, optString, fDrawNEntries,
+                outVal = tree->Draw(plotString + ">>" + histoName, cutString, optString, fDrawNEntries,
                                     fDrawFirstEntry);
                 TH3F* hh = (TH3F*)gPad->GetPrimitive(reducedHistoName);
                 if (outVal == 0) {
--- a/source/framework/core/src/TRestAnalysisPlot.cxx
+++ b/source/framework/core/src/TRestAnalysisPlot.cxx
@@ -667,15 +667,15 @@ void TRestAnalysisPlot::PlotCombinedCanvas() {
             string label = fPanels[n].label[m];
 
             size_t pos = 0;
-            label = Replace(label, "<<startTime>>", ToDateTimeString(startTime), pos);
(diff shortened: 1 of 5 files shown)


source/framework/core/src/TRestAnalysisPlot.cxx
```diff
@@ -756,10 +756,10 @@ void TRestAnalysisPlot::PlotCombinedCanvas() {
                 int outVal;
                 TString reducedHistoName = nameString + "_" + std::to_string(j);
                 TString histoName = nameString + "_" + std::to_string(j) + rangeString;
-                RESTInfo << "AnalysisTree->Draw(\"" << plotString << ">>" << histoName << "\", \""
+                RESTInfo << "AnalysisTree->Draw(\"" << plotString << "]]" << histoName << "\", \""
```
jgalan, line 759: Be careful here, it should not be replaced
```suggestion
                RESTInfo << "AnalysisTree->Draw(\"" << plotString << ">>" << histoName << "\", \""
```

jgalan:
There are other places where we use `&gt;` would it make sense to search for a different nomenclature?

```
    <addTask command="restplot-&gt;PlotCombinedCanvas()" value="ON"/>
```

lobis:
> There are other places where we use `&gt;` would it make sense to search for a different nomenclature?
> 
> ```
>     <addTask command="restplot-&gt;PlotCombinedCanvas()" value="ON"/>
> ```
> 
> We might use: `restplot.PlotCombinedCanvas()`.

I am not familiar at all with `TRestAnalysisPlots` can you explain further please? I don't understand the issue.

lobis:
I think I fixed all the problems, I don't think I broke anything with this change but I am not 100% sure as I am not familiar with `TRestAnalysisPlots`.

nkx111:
``[[xxx]]`` is a better replacing mark for me

nkx111:
> There are other places where we use `&gt;` would it make sense to search for a different nomenclature?
> 
> ```
>     <addTask command="restplot-&gt;PlotCombinedCanvas()" value="ON"/>
> ```
> 
> We might use: `restplot.PlotCombinedCanvas()`.

We can update TRestTask.cxx to support also  `restplot.PlotCombinedCanvas()`.

jgalan:
There are other places where `&gt;` is also used, for example inside the `globalCut` definition, inside `TRestAnalysisPlot`, the `condition` value can be something like `>10`.

When we really want to say greater than, how we could replace this symbol?

jgalan (approved):
Looks fine to me

jgalan:
Hi @lobis and @juanangp merging this PR made the rawlib validation pipeline fail.

https://github.com/rest-for-physics/rawlib/pull/95

It is possible to trigger the pipelines of few key libraries at the PR in the framework?

PR #368, opened 2023-02-09
Revert "use [[(]]) instead of <<(>>) for analysis plots to correct invalid xml syntax"

jgalan:
jgalan Ok: 81 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/revert-362-lobis-xml) [](https://github.com/rest-for-physics/framework/commits/revert-362-lobis-xml)

Reverts rest-for-physics/framework#362

This PR broke the rawlib pipeline as we can see at the following PR. 

https://github.com/rest-for-physics/rawlib/pull/95

@juanangp it would be easy to introduce key validation pipelines into the framework pipeline?

### DIFF
--- a/pipeline/analysisPlot/panel.xml
+++ b/pipeline/analysisPlot/panel.xml
@@ -1,14 +1,15 @@
-<panel font_size="0.06">
-    <label value="Run number : [[TRestRun::fRunNumber]]" x="0.25" y="0.9"/>
-    <label value="Run tag : [[TRestRun::fRunTag]]" x="0.25" y="0.82"/>
-    <label value="Run starts : [[startTime]]" x="0.25" y="0.74"/>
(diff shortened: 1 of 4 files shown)


lobis:
I guess this PR is just for testing right? we do not want to revert this change.

juanangp:
> I guess this PR is just for testing right? we do not want to revert this change.

Indeed, I think it would be better to identify the issue, I guess should be inside some config file.

juanangp:
> @juanangp it would be easy to introduce key validation pipelines into the framework pipeline?
I think it is possible, but it is not straightforward. I can have a look to it.

jgalan:
> Indeed, I think it would be better to identify the issue, I guess should be inside some config file.

Ok, I think this should solve the issue

https://github.com/rest-for-physics/framework/pull/370

Issue #353, opened 2022-12-15
Avoid code duplication for signal processing


juanangp:
Some functions seems duplicated within [TRestDetectorSignal](https://github.com/rest-for-physics/detectorlib/blob/master/inc/TRestDetectorSignal.h) and [TRestRawSignal](https://github.com/rest-for-physics/rawlib/blob/master/inc/TRestRawSignal.h), for instance `GetSignalSmoothed`, `GetBaseLine` and `GetBaseLineSigma`.

I propose de following:
 - Implement generic methods for signal processing that should be agnostic of the original signal event.
 - Speed up data processing by extracting different parameters such as baseline, baselineSigma, threshold integral, risetime, maxPeak and so on using a single loop (funcion).
 - The new class or namespace for signal processing should lie inside framework since I believe it should be generic for different libraries (`rawlib` `detectorlib` and `connectorslib`)

jgalan:
Probably the methods on `TRestDetectorSignal` should be just removed. The methods you mention make sense only in the domain of raw-signal processing.

Actually, even if it is named signal, the `TRestDetectorSignal` has not, or in some scenarios will not, have the properties of a signal, and many times those algorithms will not make sense. The smoothing will require a fixed binning for example.

The reason why those methods are there is because of very old times (around 2016), when we didn't have a final definition of `Short_t` array that is today `TRestRawSignalEvent`.

jgalan:
> * Speed up data processing by extracting different parameters such as baseline, baselineSigma, threshold integral, risetime, maxPeak and so on using a single loop (funcion).

This is already done inside `TRestRawSignalAnalysisProcess::ProcessEvent`. If you think it is computationally expensive we could always create a new process with reduced calculations --> `TRetRawSignalFastAnalysisProcess` that extracts the most basic observables.

jgalan:
> * Implement generic methods for signal processing that should be agnostic of the original signal event.

However, those methods are clearly to be used with a `TRetRawSignal`.

jgalan:
> * Implement generic methods for signal processing that should be agnostic of the original signal event.

Agnostic? I imagine these methods will receive as argument a `std::vector <Short_t>`, right? That is exactly what it is a `TRestRawSignal`.

jgalan:
Perhaps some of the methods inside `TRestDetectorSignal` should be moved to `TRestRawSignal`. Few methods date from long time ago when we didn't have yet a `TRestRawSignal`.

[...]

PR #379, opened 2023-02-15
WIP: Adding generic methods for signal analysis inside framework

juanangp:
New namespace added for signal analysis inside framework

### DIFF
--- a/source/framework/analysis/inc/TRestSignalAnalysis.h
+++ b/source/framework/analysis/inc/TRestSignalAnalysis.h
@@ -1,0 +1,267 @@
+/*************************************************************************
+ * This file is part of the REST software framework.                     *
+ *                                                                       *
+ * Copyright (C) 2016 GIFNA/TREX (University of Zaragoza)                *
+ * For more information see http://gifna.unizar.es/trex                  *
+ *                                                                       *
+ * REST is free software: you can redistribute it and/or modify          *
+ * it under the terms of the GNU General Public License as published by  *
+ * the Free Software Foundation, either version 3 of the License, or     *
+ * (at your option) any later version.                                   *
+ *                                                                       *
+ * REST is distributed in the hope that it will be useful,               *
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of        *
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the          *
+ * GNU General Public License for more details.                          *
+ *                                                                       *
+ * You should have a copy of the GNU General Public License along with   *
+ * REST in $REST_PATH/LICENSE.                                           *
+ * If not, see http://www.gnu.org/licenses/.                             *
+ * For the list of contributors see $REST_PATH/CREDITS.                  *
+ *************************************************************************/
+
+#include <Rtypes.h>
+#include <TMath.h>
+#include <iostream>
+#include <vector>
+
+#ifndef RestCore_TRestSignalAnalysis
+#define RestCore_TRestSignalAnalysis
+
+#include <TVector2.h>
+#include <numeric>
+
+/// This namespace define utilities (functions) to calculate different signal parameters
+namespace TRestSignalAnalysis {
+
+/// Generic functions for different calculations
+
+template <typename T>
+void CalculateBaselineAndSigmaSD(const std::vector<T>& signal, Int_t startBin, Int_t endBin,
+                                 Double_t& baseLine, Double_t& baseLineSigma) {
+    baseLine = 0;
+    baseLineSigma = 0;
+
+    int nPoints = 0;
+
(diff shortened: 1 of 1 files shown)


jgalan:
I understand it is much cleaner to create this dedicated class, however, I still believe the right place for it is at the `rawlib`. 

Thus, I suggest to rename this class to `TRestRawSignalAnalysis` and place it at rawlib.

Only in rawlib it will make sense to calculate baselines, or does it make sense in any other existing or future library?

jgalan:
If new types (other than `Short_t`) are introduced for rawsignals, those types should be added somehow also inside `rawlib`.

Please, share in case that you foresee another **use-case** where you will need a library for raw signal processing that it is outside `rawlib`.

juanangp:
> Please, share in case that you foresee another **use-case** where you will need a library for raw signal processing that it is outside `rawlib`.

I think this topic is discussed on https://github.com/rest-for-physics/framework/issues/353.
Personally, I think that signal analysis methods should be generic and doesn´t have to be linked to any library. From my understanding `rawlib` holds raw signal data but it doesn't mean that all the signal processing has to be done in this library.
For instance, I was attemting to perform the analysis of the smoothed signal which was  difficult with the current code design, this is actually implemented on  under a related PR https://github.com/rest-for-physics/connectorslib/pull/27.
In addition `detectorlib` currently holds several methods for signal processing, despite signal smoothing or baseline calculations, there are several methods such as `GetMaxLandau()`, `GetMaxGauss()` or `GetMaxAget()` that are used inside `TRestDetectorSignalToHitsProcess`.
On the other hand, I don't see any issue to perform the signal analysis on a ´TRestDetectorEvent´.

jgalan (approved):

jgalan:
Should `TRestSignalAnalysis` be renamed to something different such as `TRestTimeSignalAnalysis`, `TRestPulseShapeAnalysis` or `TRestWaveFormAnalysis`? `TRestSignal` it is quite generic and it might be used in an statistical case, such as it could be the case of using `TRestSignal` and `TRestBackground` components.

juanangp:
> Should `TRestSignalAnalysis` be renamed to something different such as `TRestTimeSignalAnalysis`, `TRestPulseShapeAnalysis` or `TRestWaveFormAnalysis`? `TRestSignal` it is quite generic and it might be used in an statistical case, such as it could be the case of using `TRestSignal` and `TRestBackground` components.

Indeed, it can be renamed since it can be misleading with new developments, I vote for `TRestPulseShapeAnalysis`

lobis (approved):

juanangp:
@lobis and @jgalan can you approve or give me feedback in the related PR? I cannot merge otherwise.
- https://github.com/rest-for-physics/rawlib/pull/100
- https://github.com/rest-for-physics/connectorslib/pull/27
- https://github.com/rest-for-physics/detectorlib/pull/75

nkx111:
If I remember correctly you cannot write the implementation of the template methods separately in the cxx file, unless you `#include` this cxx file like a .h file. I don't know why it works as the pipeline is green...

juanangp:
> If I remember correctly you cannot write the implementation of the template methods separately in the cxx file, unless you `#include` this cxx file like a .h file. I don't know why it works as the pipeline is green...

As far as I know, template specializations can be defined in the source file https://stackoverflow.com/questions/115703/storing-c-template-function-definitions-in-a-cpp-file 
I think it is cleaner to do the implementation inside the source file. The code compiles and works properly.

nkx111 (approved):

juanangp:
Note that the following submodule PR needs to be merged together with this one:
- https://github.com/rest-for-physics/rawlib/pull/100
- https://github.com/rest-for-physics/connectorslib/pull/27
- https://github.com/rest-for-physics/detectorlib/pull/75

nkx111 (approved):

source/framework/analysis/src/TRestPulseShapeAnalysis.cxx
```diff
@@ -0,0 +1,667 @@
...
+/// we find an overshoot, being the baseline not returning to zero (or its
+/// original value) at the signal tail.
+///
+/// \param baseLine value of the signal baseline calculated beforehand
+///
+/// \param baseLineSigma value of the baseline fluctuation calculated beforehand
+///
+
+template <typename T>
+std::vector<std::pair<Float_t, Float_t> > TRestPulseShapeAnalysis::GetPointsOverThreshold(
```
nkx111, line 280: Could return `std::vector<std::pair<Int_t, T> >`?

juanangp: I guess this is related to https://github.com/rest-for-physics/rawlib/pull/100#discussion_r1203847342 
I will try to implement a more generic method.

PR #382, opened 2023-02-26
New metadata class to merge TRestRun files

juanangp:
New metadata class TRestRunMerger meant to merge different TRestRun files, merging of the files can be performed using rml config files or macros.

### DIFF
--- a/macros/REST_MergeFiles.C
+++ b/macros/REST_MergeFiles.C
@@ -1,32 +1,15 @@
-#include "TFileMerger.h"
-#include "TRestTask.h"
-
-#ifndef RESTTask_MergeFiles
-#define RESTTask_MergeFiles
+#include "TRestRunMerger.h"
 
 //*******************************************************************************************************
 //***
-//*** Your HELP is needed to verify, validate and document this macro
-//*** This macro might need update/revision.
+//*** Updated REST_MergeFiles macro with the new TRestRunMerger class
+//*** See TRestRunMerger for more details about this implementation
 //***
 //*******************************************************************************************************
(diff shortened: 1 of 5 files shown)


jgalan:
Why would we need to merge files? Each REST file has a unique TRestRun entry that identifies the data file. That's why I find a bit strange merging TRestRun files. Although we did this in the past ...

The data merging can be done at the TRestDataSet level, where the metadata of the TRestDataSet allows to univocally identify the file list used to generate the dataset from a persistent read-only storage.

juanangp:
> The data merging can be done at the TRestDataSet level, where the metadata of the TRestDataSet allows to univocally identify the file list used to generate the analysis dataset from a persistent read-only storage.

I am trying to perform the energy calibration of the data, this is the workflow that I have in mind:
1. Data selection to avoid large files, this is done using TRestCut
2. Merging of the files, now is possible using TRestRunMerger
3. Perform the calibration and store the calibration constants and the converted energy in a TRestRun
4. Perform a further analysis of the data.

So, in summary I would like to have a TRestRun with the calibration energy, but the calibration is split along several files, that's why I think it will be useful to perform the merging of TRestRuns. Also, note that the merging of the files is only performed among the files with the same run number.
The problem of TRestDataSet is that is read-only and I cannot save the calibration constants or the calibrated energy inside the files. So, I think that TRestDataSet can be used in step 4 but not in the steps before.
Otherwise, can you point me to a different workflow?

lobis (approved):

lobis:
It's interesting that the `clang-format` failed even after the PR bot made the formatting changes... 🤔

PR #385, opened 2023-03-03
Updating validation file of PandaX-III

jgalan:
jgalan Ok: 0 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/jgalan_smear_review2) [](https://github.com/rest-for-physics/framework/commits/jgalan_smear_review2)

Due to update of TRestDetectorHitsSmearingProcess at rest-for-physics/detectorlib#72


jgalan:
@nkx I had to update the validation file for PandaX-III because I updated the way it is initialised the random generator. I am not 100% sure of the reason why this is like that, but my guess is because now the `random` generator is initialised at `InitProcess` and there are two calls to `new TRandom`. Thus, the first event in the final processing chain is executed just after initialisation. In the previous version I think it was initialised inside `InitFromConfigFile` and thus, the test loop was executing the first random call.

lobis:
Perhaps these files are starting to become too big? The main framework repository is already at `105.69 MB` size. This would increase the total size by ~3%.

jgalan:
> Perhaps these files are starting to become too big? The main framework repository is already at `105.69 MB` size. This would increase the total size by ~3%.

Ups, in fact I was not supposed to upload the Mb file. Only the 13kbytes one, I didn't notice, thanks! I will just push force.

jgalan:
Fixed!

juanangp (approved):

nkx111:
Looks fine to me

nkx111 (approved):

Issue #390, opened 2023-03-10
Bug when processing files with a single event


jgalan:
It seems that we are in troubles when the original file contains only one input event.

We get a segmentation fault inside `TRestRun::GetNextEvent` since `fInputEvent` seems to be a nullptr.

The problem can be reproduced using run 646.

```
restManager --v 4 --c ShieldingContamination.rml --f /tmp/output_646.root
```

It cannot be reproduced using run 953 that contains 3 events.

The attached files allow to reproduce the problem.

[processing.tgz](https://github.com/rest-for-physics/framework/files/10941762/processing.tgz)

Please @nkx111 could you have a look to this? It is kind of urgent and there are many things you introduced in `TRestProcessRunner` with threads that are a black box to us.

Thank you!

PR #393, opened 2023-04-18
Fix bug when processing files with a single event

nkx111:
nkx111 Ok: 40 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/nkx111-single-event-file) [](https://github.com/rest-for-physics/framework/commits/nkx111-single-event-file)

Issue #390.
The bug comes from when finished doing test run(before the actual event processing), REST will push-forward to the next event. Since there is only one event in the file, find the next non-existing event will actually trigger finish reading. Therefore the later processes will get null input event, and cause problems. This fix it to prevent push-forward of event entries during test run, if test run succeeds.


jgalan:
Nice! I guess some pipelines are failing because it is not the same event that it is being evaluated?

juanangp:
> Nice! I guess some pipelines are failing because it is not the same event that it is being evaluated?

Still don't understand why some pipelines are failing, if you are checking a different event why only one observable has a different value e.g. on the TREX-DM validation pipeline? I would expect several observables failing...

jgalan:
> > Nice! I guess some pipelines are failing because it is not the same event that it is being evaluated?
> 
> Still don't understand why some pipelines are failing, if you are checking a different event why only one observable has a different value e.g. on the TREX-DM validation pipeline? I would expect several observables failing...

The observable that it is failing is `rate_MeanRate_InHz` this observable does not only depend on the specific entry being evaluated but it takes the timestamp value of a different event (shifted by 10 events, for example) to calculate the mean value. If there is an extra event ahead this value could be affected. This is a guess.

For the ray-tracing case, all the values are changing, so I guess it is an issue affecting the seed by the fact that there is one more event entry? @nkx111 ?

juanangp:
> rate_MeanRate_InHz

Then this observable doesn't make much sense to me, in general we should have one observable per event, the rate can be computed later on...

nkx111:
> The observable that it is failing is rate_MeanRate_InHz

I looked into each timestamp of the first 10 events from the pipeline. It changes after this PR. I suspect that it is related somehow with the bug in TRestRawMultiFEMINOSToSignalProcess we already recorded. https://github.com/rest-for-physics/rawlib/issues/36

Before the PR:
```
$test run: 1582141590.00456

$event 0: 1582141590.00629
$event 1: 1582141590.00456
$event 2: 1582141590.00629
$event 3: 1582141590.00627
$event 4: 1582141590.0068
$event 5: 1582141590.00771
$event 6: 1582141590.00961
...
```

With this PR:
```
$test run: 1582141590.00456

$event 0: 1582141590.00456
$event 1: 1582141590.00456
$event 2: 1582141590.00629
$event 3: 1582141590.00627
$event 4: 1582141590.0068
$event 5: 1582141590.00771
$event 6: 1582141590.00961
```

The assumed "correct behavior" is that the first two events from `TRestRawMultiFEMINOSToSignalProcess` is identical, which is exactly the case with this PR. Before this PR, however, with additional "get-next-event" operation called after test run, event id 0 mysteriously turns into event id 2. 

Anyway, despite the first few values of `rate_MeanRate_InHz`, this observable is correct and meaningful. I think we can just update the checks. In future when we fix the bug in `TRestRawMultiFEMINOSToSignalProcess`, the validation file shall be updated again.

nkx111:
> For the ray-tracing case, all the values are changing, so I guess it is an issue affecting the seed by the fact that there is a different looping scheme?

Yes, I think this is the case. Previously we are calling `TRestAxionGeneratorProcess::ProcessEvent()` once more than this PR. So the random value gets changed.

jgalan:
> > rate_MeanRate_InHz
> 
> Then this observable doesn't make much sense to me, in general we should have one observable per event, the rate can be computed later on...

Yes, that might be a good choice. But does that mean that we need to move `https://sultan.unizar.es/rest/classTRestEventRateAnalysisProcess.html` to legacy?

juanangp (approved):

jgalan (approved):

PR #394, opened 2023-04-18
clear-install-dir is performed before add-subdir

nkx111:
With the latest version of cmake I found the binary files are not appearing after installation. I found it is caused by clear-install-dir action which is executed now after subdir installation. So I move forward the lines for clearing the directory.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -212,6 +212,16 @@ include(Testing)
 
 # Start compile
 include(MacroRootDict)
+
+# Clear the install dir
+if (NOT DEFINED INSTALL_CLEARDIR)
+    set(INSTALL_CLEARDIR ON)
+endif ()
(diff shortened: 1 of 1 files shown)


jgalan:
Ah! Maybe this PR fixes some problems detected on MacOs, perhaps it was related to the cmake version.

https://rest-forum.unizar.es/t/macos-installation-issue-missing-files-headers-bin/573

This was happening only with restG4 I believe. 

@heucheld @acorsi

juanangp (approved):
I don't think we should remove the install dir every time that we do `make install`, can you please create an issue about this to follow up the proper handling of removing the install dir

jgalan:
> I don't think we should remove the install dir every time that we do `make install`, can you please create an issue about this to follow up the proper handling of removing the install dir

Just created

PR #407, opened 2023-05-03
mpreal.h has been removed

raulena333:
mpreal.h has been removed because it caused conflicts with some versions of MPFR libraries.

This compatibility issue seemed to be originated by the `mpreal.h` header present in the framework.

The mpfr.h is included inside the linux package `libmpfrc++-dev`


jgalan:
Hi @lobis, it would be better to include the library `libmpfrc++-dev` inside the docker. Because it creates problems when the `mpfr-dev` library does not match the same version as the hard coded header inside the framework.

It would be an issue to add that package to the docker?

jgalan:
> It should not be a problem (I assume it's installed as `apt-get install libmpfrc++-dev` in Ubuntu).

Yes, it is installed this way

lobis:
> > It should not be a problem (I assume it's installed as `apt-get install libmpfrc++-dev` in Ubuntu).
> 
> Yes, it is installed this way.
> 
> It has been added in the recent docker update?

Yes it has been added, now we need to merge all PRs to update the image:

https://github.com/rest-for-physics/framework/issues/413

PR #409, opened 2023-05-04
TRestAnalysisTree fixing observable estimators 

jgalan:
jgalan Medium: 115 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/jgalan_fix_restG4) [](https://github.com/rest-for-physics/framework/commits/jgalan_fix_restG4)

TRestAnalysisTree::GetObservableAverage. Changing the way average is calculated

Other similar methods `GetObservableMax/Min/RMS` have been updated too

### DIFF
--- a/source/framework/core/inc/TRestAnalysisTree.h
+++ b/source/framework/core/inc/TRestAnalysisTree.h
@@ -279,25 +279,24 @@ class TRestAnalysisTree : public TTree {
     void EnableQuickObservableValueSetting();
     void DisableQuickObservableValueSetting();
 
-    Double_t GetIntegral(const TString& obsName, Double_t xLow = -1, Double_t xHigh = -1,
-                         Int_t nBins = 1000) {
-        return GetObservableIntegral(obsName, xLow, xHigh, nBins);
+    Double_t GetIntegral(const TString& obsName, Double_t xLow = -1, Double_t xHigh = -1) {
+        return GetObservableIntegral(obsName, xLow, xHigh);
     }
 
     Double_t GetAverage(const TString& obsName, Double_t xLow = -1, Double_t xHigh = -1) {
         return GetObservableAverage(obsName, xLow, xHigh);
     }
 
-    Double_t GetRMS(const TString& obsName, Double_t xLow = -1, Double_t xHigh = -1, Int_t nBins = 1000) {
-        return GetObservableRMS(obsName, xLow, xHigh, nBins);
+    Double_t GetRMS(const TString& obsName, Double_t xLow = -1, Double_t xHigh = -1) {
+        return GetObservableRMS(obsName, xLow, xHigh);
     }
 
-    Double_t GetMinimum(const TString& obsName, Double_t xLow = -1, Double_t xHigh = -1, Int_t nBins = 1000) {
-        return GetObservableMinimum(obsName, xLow, xHigh, nBins);
+    Double_t GetMinimum(const TString& obsName, Double_t xLow = -1, Double_t xHigh = -1) {
+        return GetObservableMinimum(obsName, xLow, xHigh);
     }
 
-    Double_t GetMaximum(const TString& obsName, Double_t xLow = -1, Double_t xHigh = -1, Int_t nBins = 1000) {
-        return GetObservableMaximum(obsName, xLow, xHigh, nBins);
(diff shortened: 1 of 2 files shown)


jgalan:
This fixes the pipeline, still I don't understand what is wrong with the previous implementation.

I don't like to iterate manually on the entries. Specially because `TRestAnalysisTree` might be pointing to a given entry that is the same as the entry in `TRestRun` event tree.

juanangp:
> This fixes the pipeline, still I don't understand what is wrong with the previous implementation.
> 
> I don't like to iterate manually on the entries inside the analysis tree. Specially because `TRestAnalysisTree` might be pointing to a given entry that is the same as the entry in `TRestRun` event tree.
> 
> This might be an issue if we call `GetObservableAverage` in between requesting an entry using `TRestRun::GetEntry` and the access to the information.

You can try with `RDataFrame`
```
RDataFrame df (*myAnalysisTree);
double mean = *df.Mean(obsName);
```

jgalan:
> > This fixes the pipeline, still I don't understand what is wrong with the previous implementation.
> > I don't like to iterate manually on the entries inside the analysis tree. Specially because `TRestAnalysisTree` might be pointing to a given entry that is the same as the entry in `TRestRun` event tree.
> > This might be an issue if we call `GetObservableAverage` in between requesting an entry using `TRestRun::GetEntry` and the access to the information.
> 
> You can try with `RDataFrame`
> 
> ```
> RDataFrame df (*myAnalysisTree);
> double mean = *df.Mean(obsName);
> ```

Using `RDataFrame` is much slower.

```
Processing Validate.C...
Entries: h1: 1984, h2: 7432, h3: 9430
====== Theta average ====
Elapsed time in nanoseconds: 1026357322 ns
Elapsed time in microseconds: 1026357 µs
Elapsed time in milliseconds: 1026 ms
====== Phi average ====
Elapsed time in nanoseconds: 144700115 ns
Elapsed time in microseconds: 144700 µs
Elapsed time in milliseconds: 144 ms
```

Using the direct tree access:

```
Processing Validate.C...
Entries: h1: 1984, h2: 7432, h3: 9430
====== Theta average ====
Elapsed time in nanoseconds: 618828 ns
Elapsed time in microseconds: 618 µs
Elapsed time in milliseconds: 0 ms
====== Phi average ====
Elapsed time in nanoseconds: 613547 ns
Elapsed time in microseconds: 613 µs
Elapsed time in milliseconds: 0 ms
```

I thought that would happen only the first time I called to the `df.Mean` method, since the first time it would probably require to initialise the data frame. But the chronometer for the second call gives around 150ms for both!

So the DataFrame method is bullshit?!?!

juanangp:
> I thought that would happen only the first time I called to the `df.Mean` method, since the first time it would probably require to initialise the data frame. But the chronometer for the second round just gives around 150ms for both!
> 
> So the DataFrame method is bullshit?!?!

It should happen everytime you call the method because you generate a dataset.
I am not surprised of the time delay, it takes a long time to generate a dataset, I would not say it is bullshit, it is just not designed for this user case.

jgalan:
> > I thought that would happen only the first time I called to the `df.Mean` method, since the first time it would probably require to initialise the data frame. But the chronometer for the second round just gives around 150ms for both!
> > So the DataFrame method is bullshit?!?!
> 
> It should happen everytime you call the method because you generate a dataset. I am not surprised of the time delay, it takes a long time to generate a dataset, I would not say it is bullshit, it is just not designed for this user case.

Then, what it is the proper **fast** way to calculate the mean using the RDataFrame?

jgalan:
Could you approve this PR so that we fix pipelines?

juanangp:
> Then, what it is the proper **fast** way to calculate the mean using the RDataFrame?

There is not fast way, it is what it is, I guess is not using multithreading but anyhow I don't think it will improve since it is a lazy method

juanangp (approved):

jgalan:
> > Then, what it is the proper **fast** way to calculate the mean using the RDataFrame?
> 
> There is not fast way, it is what it is, I guess is not using multithreading but anyhow I don't think it will improve since it is a lazy method

Ok, so that means the `Mean` method from RDataFrame should not be used.

lobis (approved):

PR #418, opened 2023-05-09
MacroRootDict.cmake. Adding install exception for MacOs

jgalan:
jgalan Ok: 6 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/jgalan_macos_fix2) [](https://github.com/rest-for-physics/framework/commits/jgalan_macos_fix2)

See also https://github.com/rest-for-physics/restG4/pull/109 and https://github.com/rest-for-physics/axionlib/pull/64

PR #419, opened 2023-05-11
Fixing MacOs libraries identification

jgalan:
jgalan Ok: 4 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/jgalan_fix_macos_libs) [](https://github.com/rest-for-physics/framework/commits/jgalan_fix_macos_libs)

This PR solves the problem of identification of libraries that has been discussed in the Forum: 

https://rest-forum.unizar.es/t/macos-installation-issue-missing-files-headers-bin/573/13.

If this PR solves the problem (it seems to solve the problem in my Mac) then we can just merge this PR, and avoid merging the following PRs.

https://github.com/rest-for-physics/framework/pull/418
https://github.com/rest-for-physics/restG4/pull/109
https://github.com/rest-for-physics/axionlib/pull/64

which should be simply closed.


cmake/thisREST.cmake
```diff
@@ -128,6 +130,8 @@ export REST_GARFIELD_INCLUDE=${Garfield_INCLUDE_DIRS}
 export REST_GARFIELD_LIB=${Garfield_LIBRARIES}
 export PATH=\\\$REST_PATH/bin:\\\$_PATH
 export LD_LIBRARY_PATH=\\\$REST_PATH/lib:\\\$_LD_LIBRARY_PATH
+# DYLD_LIBRARY_PATH is required by MacOs
+export DYLD_LIBRARY_PATH=\\\$REST_PATH/lib:\\\$LD_LIBRARY_PATH${g4LibPath}
```
lobis, line 134:
```suggestion
export DYLD_LIBRARY_PATH=\\\$REST_PATH/lib:\\\$DYLD_LIBRARY_PATH${g4LibPath}
```

Otherwise the command will overwrite the `DYLD_LIBRARY_PATH` the user might have, which can cause conflicts (some other libraries might not be found, etc.).

jgalan: Yes, but, `DYLD_LIBRARY_PATH` was not defined at all. All libraries usually take of setting `LD_LIBRARY_PATH`, so I thought it would be better to initialise `DYLD_LIBRARY_PATH` with all those libraries that are already defined in `LD_LIBRARY_PATH` to make sure `DYLD_LIBRARY_PATH` contains all the necessary libraries.

jgalan: For example, if I do `source /path/to/root/install/bin/thisroot.sh`. The `LD_LIBRARY_PATH` is set, but not `DYLD_LIBRARY_PATH`.

```
source ~/apps/root-6.26.10/install/bin/thisroot.sh 
Javiers-MBP:~ jgalan$ env | grep LD
LD_LIBRARY_PATH=/Users/jgalan/apps/root-6.26.10/install/lib
Javiers-MBP:~ jgalan$ 
```

lobis: That is interesting, because taking a quick loot at `thisroot.sh` it seems that it would define `DYLD_LIBRARY_PATH`. It seems that root takes the approach of defining both `LD` and `DYLD` regardless of system (but in practise it's not being defined for some reason?).

My problem with line https://github.com/rest-for-physics/framework/pull/419/files#diff-558c59afe63a1a4fdcccf970d21d2c543ce968473e99eabe80e1050c9505c184R134 is that it overwrites the `DYLD_LIBRARY_PATH` variable, what if the user doesn't want this variable overwritten because it already contains some useful value?

jgalan: It will be already in `LD_LIBRARY_PATH`

jgalan: As you see also the `thisroot.sh` script is filling `LD_LIBRARY_PATH` so when we do `DYLD_LIBRARY_PATH=$LD_LIBRARY_PATH` it solves the problem. It is strange, but `thisroot.sh` is not setting `DYLD_LIBRARY_PATH`. It is a MacOs obscure thing I believe.

jgalan: Perhaps @heucheld could confirm that a clean `master` REST installation + this PR solves the problem?

heucheld: For me it also looks like a very obscure and weird MacOS specific problem/thing. 
**I can confirm: With a clean master REST installation and the fix on branch jgalan_fix_macos_libs for the framework, everything works well and as expected on my two local MacOS systems. No need to copy any libraries anymore.**

heucheld (approved):

lobis (approved):

PR #427, opened 2023-05-24
Fix on TRestMetadata and TRestDataSet add-ons

jgalan:
jgalan Medium: 165 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/jgalan_minor_fix) [](https://github.com/rest-for-physics/framework/commits/jgalan_minor_fix)

`TRestMetadata::GetDataMemberValues` minor fix on TVector2,3 output component

Before this PR (for the case of TVector2,3), the following output was generated when invoking `GetMetadataMember`.

```
root [7] run0->GetMetadataMember("axionGen::fEnergyRange[0]")
(std::string) "(0.05"
root [8] run0->GetMetadataMember("axionGen::fEnergyRange[1]")
(std::string) "10)"
```

After this PR, parenthesis are removed from the output.

```
root [7] run0->GetMetadataMember("axionGen::fEnergyRange[0]")
(std::string) "0.05"
root [8] run0->GetMetadataMember("axionGen::fEnergyRange[1]")
(std::string) "10"
```

Added also:

- `REST_Generate_DataSets` macro that will allow to export several datasets from a given list of dataset names.
- Added new filter condition `equalsTo` that allows to evaluate numerically if a datamember is equal to the filter. If it is not equal, then the run will be rejected and thus not included in the dataset.
- Added method `TRestDataSet::Define` that allows to create a new column using the values of relevant quantities.
- Added method `TRestDataSet::Import( std::vector <std::string> fNames )` allowing to combine datasets into one single instance. If saved to disk the metadata member `fMergedDataSet` will be `true` in order to identify that it is a non-pure original dataset.

For the moment metadata filter conditions do not support units. See issue: https://github.com/rest-for-physics/framework/issues/429

### DIFF
--- a/macros/REST_GenerateDataSet.C
+++ b/macros/REST_GenerateDataSet.C
@@ -22,7 +22,7 @@ Int_t REST_GenerateDataSet(std::string inputRML, std::string datasets) {
         std::cout << "Set : " << set << std::endl;
         TRestDataSet d(inputRML.c_str(), set.c_str());
         d.GenerateDataSet();
-        d.Export("dataset_" + set + ".root");
+        d.Export("Dataset_" + set + ".root");
     }
     return 0;
 }
--- a/macros/REST_GenerateDataSet.C
+++ b/macros/REST_GenerateDataSet.C
@@ -1,0 +1,32 @@
+#include "TRestTask.h"
+#include "TRestDataSet.h"
+
+#ifndef RestTask_GenerateDataSet
+#define RestTask_GenerateDataSet
+
+//*******************************************************************************************************
+//*** Description: This macro will launch the generation of datasets defined
+//*** inside a particular RML file `datasets.rml` that contains the dataset
+//*** definitions. The second argument will allow to specify the datasets
+//*** to be generated from the existing ones inside `dataset.rml`.
+//***
+//*** --------------
+//*** Usage: restManager GenerateDataSet datasets.rml set1,set2,set3
+//***
+//*******************************************************************************************************
+
+Int_t REST_GenerateDataSet( std::string inputRML, std::string datasets )
+{
+
+	std::vector<std::string> sets = REST_StringHelper::Split( datasets, "," );
+
+	for( const auto &set: sets ) {
+		std::cout << "Set : " << set << std::endl;
+		TRestDataSet d( inputRML.c_str(), set.c_str() ); 
+		d.GenerateDataSet();
+		d.Export( "dataset_" + set + ".root" );
+	}
+	return 0;
+
+}
+#endif
--- a/macros/REST_GenerateDataSet.C
+++ b/macros/REST_GenerateDataSet.C
@@ -1,5 +1,5 @@
-#include "TRestTask.h"
 #include "TRestDataSet.h"
+#include "TRestTask.h"
 
 #ifndef RestTask_GenerateDataSet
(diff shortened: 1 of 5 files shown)


macros/REST_GenerateDataSet.C
```diff
@@ -0,0 +1,29 @@
...
+//*** inside a particular RML file `datasets.rml` that contains the dataset
+//*** definitions. The second argument will allow to specify the datasets
+//*** to be generated from the existing ones inside `dataset.rml`.
+//***
+//*** --------------
+//*** Usage: restManager GenerateDataSet datasets.rml set1,set2,set3
+//***
+//*******************************************************************************************************
+
+Int_t REST_GenerateDataSet(std::string inputRML, std::string datasets) {
```
juanangp, line 18: From my experience, I think it would be more useful to generate a dataset using the file pattern. E.g.
```
Int_t REST_GenerateDataSet(const std::string &inputRML, const std::string &datasetName, const std::string &filePattern, const std::string &outputFileName) {
    TRestDataSet d(inputRML.c_str(), datasetName.c_str());
    d.SetFilePattern(filePattern);
    d.GenerateDataSet();
    d.Export(outputFileName);
    return 0;
}
```

jgalan: The idea is to generate several datasets at once, then the output file name takes the name of the dataset with the `dataset_` root.

jgalan: On top of that is that each dataset has a different `filePattern`.

I am using it as follows with the following file: https://github.com/iaxo/iaxo-simulations/blob/master/montecarlo/signal/sensitivity/dataset.rml

```
restManager GenerateDataSet BabyIAXO_XMM_mm_Vacuum_0meV,BabyIAXO_XMM_mm_Vacuum_1meV,BabyIAXO_XMM_mm_Vacuum_10meV,BabyIAXO_XMM_mm_Vacuum_100meV,BabyIAXO_XMM_mm_Vacuum_1eV
```

To generate:

Screenshot 2023-05-25 at 12 02 53

juanangp: Then I would suggest to start the dataset with `D` contrary to the TRestRuns which typically starts with `R`

jgalan: Ok, I could start with upper case letter, this was my default choice in the past. I think @lobis prefers start with lowercase.

lobis: Right now I don't recall stating any preference, but yes in general I use lowercase names for all file names (easier to type in my opinion) but it's up to personal preference 😃

lobis, line 18:
```suggestion
Int_t REST_GenerateDataSet(const std::string& inputRML, const std::string& datasets) {
```

cmargalejo (approved):

jgalan:
I am going a need an `Import` method that allows me to combine different datasets to be loaded in `TRestComponent`. Thus, I have implemented a new `Import` method allowing to receive several ROOT files as argument. There will be a datamember `fMergedDataSet` that keeps track if this method was used to intialize the dataset.

I also added a new `TRestDataSet::Define` method that allows to generate a new column using/replacing the values of relevant quantities.

Any feedback about their implementation? @juanangp

PR #428, opened 2023-05-24
TRestHits: error in gaussian fit is now defined by the user

cmargalejo:
cmargalejo Ok: 19 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/cris_hitsGaussError) [](https://github.com/rest-for-physics/framework/commits/cris_hitsGaussError)

This PR is required by PR https://github.com/rest-for-physics/detectorlib/pull/82
The hard coded numbers are now arguments of the functions.

### DIFF
--- a/source/framework/core/inc/TRestHits.h
+++ b/source/framework/core/inc/TRestHits.h
@@ -125,9 +125,9 @@ class TRestHits {
     Double_t GetSkewXY() const;
     Double_t GetSkewZ() const;
 
-    Double_t GetGaussSigmaX();
-    Double_t GetGaussSigmaY();
-    Double_t GetGaussSigmaZ();
+    Double_t GetGaussSigmaX(Double_t error = 150.0);
+    Double_t GetGaussSigmaY(Double_t error = 150.0);
+    Double_t GetGaussSigmaZ(Double_t error = 150.0);
 
     Double_t GetEnergyX() const;
     Double_t GetEnergyY() const;
--- a/source/framework/core/inc/TRestHits.h
+++ b/source/framework/core/inc/TRestHits.h
@@ -125,9 +125,9 @@ class TRestHits {
     Double_t GetSkewXY() const;
     Double_t GetSkewZ() const;
 
-    Double_t GetGaussSigmaX(Double_t error = 150.0);
-    Double_t GetGaussSigmaY(Double_t error = 150.0);
-    Double_t GetGaussSigmaZ(Double_t error = 150.0);
+    Double_t GetGaussSigmaX(Double_t error = 150.0, nHitsMin = 100000);
+    Double_t GetGaussSigmaY(Double_t error = 150.0, nHitsMin = 100000);
+    Double_t GetGaussSigmaZ(Double_t error = 150.0, nHitsMin = 100000);
 
     Double_t GetEnergyX() const;
     Double_t GetEnergyY() const;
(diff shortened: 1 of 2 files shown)


jgalan:
It seems that `doHitsCorrection` could be a variable that is also controlled by argument? Actually it seems hardcoded to be true.

Therefore, a potential prototyping is:

 ```
Double_t TRestHits::GetGaussSigmaZ(Bool_t doHitsCorrection = true, Double_t error = 150)
```

**Comment by jgalan** [comment]
I don't see why we need two counters here 
```
for (unsigned int n = 0; n < GetNumberOfHits(); k++, n++) {`
```

**Comment by cmargalejo** [comment]
> It seems that `doHitsCorrection` could be a variable that is also controlled by argument? Actually it seems hardcoded to be true.
> 
> Therefore, a potential prototyping is:
> 
> ```
> Double_t TRestHits::GetGaussSigmaZ(Bool_t doHitsCorrection = true, Double_t error = 150)
> ```

In the beginning I thought about leaving the option to the user, mainly thinking about data in which all the channels are saved. But then I found that it is usually better to just use it always. However, I left it there as a hint in case at some point we decide it should be controled by the user, or even if we want to activate it only if the number of hits is smaller than some number, e.g., like in the comment in line 778:  `// bool doHitCorrection = nHits <= 18; //in case we want to apply it only to the smaller events`.

For now I think it is fine as it is, but I also don't mind changing it if you think it's better.

**Review by jgalan — APPROVED** [review]

**Comment by jgalan** [comment]
> > It seems that `doHitsCorrection` could be a variable that is also controlled by argument? Actually it seems hardcoded to be true.
> > Therefore, a potential prototyping is:
> > ```
> > Double_t TRestHits::GetGaussSigmaZ(Bool_t doHitsCorrection = true, Double_t error = 150)
> > ```
> 
> In the beginning I thought about leaving the option to the user, mainly thinking about data in which all the channels are saved. But then I found that it is usually better to just use it always. However, I left it there as a hint in case at some point we decide it should be controled by the user, or even if we want to activate it only if the number of hits is smaller than some number, e.g., like in the comment in line 778: `// bool doHitCorrection = nHits <= 18; //in case we want to apply it only to the smaller events`.
> 
> For now I think it is fine as it is, but I also don't mind changing it if you think it's better.

Perhaps then what you want to pass by argument is the number of hits from which you consider the event has few hits, and correction should be applied.

```
Double_t TRestHits::GetGaussSigmaZ(Double_t error = 150, Int_t nHits = 1000000)
```

The correction applies to events with hits below `nHits`. So if you want, it is possible to add a new parameter at the process that defines for which event size we start to do the correction. The default `nHits=1000000` is that correction applies to any event. You could make 1000000 the biggest int number.

**Comment by cmargalejo** [comment]
> I don't see why we need two counters here
> 
> ```
> for (unsigned int n = 0; n < GetNumberOfHits(); k++, n++) {`
> ```
> 
> Both take always the same value inside the loop, so `n++` would be enough?

No, they don't take the same value. ```n``` starts from 0 and ```k``` comes from:
```
if (doHitCorrection) {
            nAdd = 2;
        }
Int_t nElems = nHits + nAdd;
vector<Double_t> x(nElems), y(nElems), ex(nElems), ey(nElems);
Int_t k = nAdd / 2;
´´´

where nAdd is the number of hits we add to the event.

KonradAltenmueller (approved):

PR #432, opened 2023-06-01
TRestDataSet now allows new user defined columns at the dataset initialisation stage.

jgalan:
jgalan Ok: 69 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/jgalan-dataset-upgrade) [](https://github.com/rest-for-physics/framework/commits/jgalan-dataset-upgrade)

Added possibility to add new user defined columns during the dataset generation


source/framework/core/inc/TRestDataSet.h
```diff
@@ -101,6 +101,12 @@ class TRestDataSet : public TRestMetadata {
     /// The list of dataset files imported
     std::vector<std::string> fImportedFiles;  //<
 
+    /// A list of new columns added during the dataset generation
+    std::vector<std::string> fColumnNames;
+
+    /// The list of expressions correspoding to each fColumnNames element
+    std::vector<std::string> fColumnExpressions;
+
```
juanangp, line 109: I would change it to:
´´´
std::vector<std:pair<std::string, std::string >> fColumnNameExpressions;
´´´

jgalan: I am trying to use `std::pair`, but when I try to iterate:

```
    for( const auto &[cName,cExpression]: fColumnNameExpressions )
        finalList.emplace_back(cName);
        fDataSet.Define(cName, cExpression);
    }
```

it gives me error 

```
/home/jgalan/rest-framework/source/framework/core/src/TRestDataSet.cxx:355:25: error: ‘cName’ was not declared in this scope
         fDataSet.Define(cName, cExpression);
```

jgalan: The following is also not working for me:

```
    for( const auto &column: fColumnNameExpressions )
        finalList.emplace_back(column.first);
        fDataSet.Define(column.first, column.second);
    }
```

juanangp: This should work:
´´´
    for( const auto &[cName, cExpression] : fColumnNameExpressions ) {
        finalList.emplace_back(cName);
        fDataSet.Define(cName, cExpression);
    }
´´´
You forget the `}` ?

jgalan: ok

source/framework/core/src/TRestDataSet.cxx
```diff
@@ -330,6 +349,12 @@ void TRestDataSet::GenerateDataSet() {
 
     fDataSet = MakeCut(fCut);
 
+    // Adding new user columns added to the dataset
+    for (const auto& [cName, cExpression] : fColumnNameExpressions) {
+        finalList.emplace_back(cName);
+        this->Define(cName, cExpression);
```
juanangp, line 355: I would change the behaviour of the `Define` function to something similar to `MakeCut` e.g:
```
 fDataSet = Define(cName, cExpression);
```
And in the `Define` function
```
ROOT::RDF::RNode TRestDataSet::Define(const std::string& columnName, const std::string& formula) {
  auto df = fDataSet;  
  std::string evalFormula = formula;
    for (auto const& [name, properties] : fQuantity)
        evalFormula =
            REST_StringHelper::Replace(evalFormula, name, DoubleToString(properties.value, "%12.10e"));
    df = df.Define(columnName, evalFormula);
    return df;
}
```
Then it should be possible to define a new observable but you should be capable to use or store de dataframe somewhere else. I think is more useful this way.

jgalan: Actually the `RDataFrame` is already being returned as it is done in `MakeCut` the only difference I see is that I just do not do `fDataSet = this->Define( ...)` since I know that `fDataSet` has been already updated inside.

jgalan: Also, I do not create a temporal `df` inside the method, it is this necessary?

jgalan: Oh what you want is that the internal `fDataSet` is not modified at all, then the returned contains the new column?

What if I want to add 3 new columns that use relevant quantities?

jgalan: Ok, what you want is to have the possibility to add a new column in a returned `RDataFrame`, but  this limits the operation to just 1-column, isn't? I mean, if you want to use the method outside the `dataset` you will only be able to use relevant quantities once.

Perhaps we should give as argument then the `data frame` where we will add the new evaluated column?

jgalan: I guess that becomes complicated

jgalan: Then I just change it to `AddColumn`

jgalan: Ok, finally I just leave it as `MakeCut` philosophy to be consistent.

juanangp (approved):

cmargalejo (approved):

Issue #358, opened 2022-12-28
Processes streamer input/output requires deep review


jgalan:
See root-forum discussion: https://root-forum.cern.ch/t/checkbytecount-and-streamer-error/52716/58

This problem was addressed at issue #348 and #357.

A quick patch solution was added for the case when no input file is added, as it is the case for axionlib ray-tracing. However, problems still exist in trex-dm and geant4 event processing.

For example, in `pipeline/trex` validation, after `03_hits.rml`, the processes `TRestDetectorSignalToHitsProcess` and `TRestDetectorHitsAnalysisProcess` are not present in `TFile::ShowStreamerInfo` inside `Hits_01928.root`.

Processes are not properly added to the streamer, and thus this will prevent schema evolution to be working properly.

nkx111:
Hi, I think I have located the bug. It is related to a strange behavior of ROOT streamer. If we open a file with "update" mode but do nothing, then the streamerinfo from the file will somehow lost during the next update and write. I can reproduce the problem with the following code in `restRoot`:

```c++
// create and write a file with metadata, then close the file
TFile *f = new TFile("abc.root", "recreate");
TRestRawMultiFEMINOSToSignalProcess *p = new TRestRawMultiFEMINOSToSignalProcess();
p->Write();
f->Write();
f->ShowStreamerInfo();
f->Close();
// reopen the file with "update" mode but do nothing
TFile *fff = new TFile("abc.root", "update");
fff->Close();
delete fff;
// open the file again and write another metadata. After this operation, the streamerinfo of the previous metadata gets lost
TFile *ff = new TFile("abc.root", "update");
TRestAxionSolarFlux *fl = new TRestAxionSolarFlux("fluxes.rml", "LennertHoofPrimakoff");
fl->Write();
ff->Write();
ff->ShowStreamerInfo()
```

I am still thinking how to work around with it

jgalan:
Perhaps this should be reported to the root-forum?

PR #434, opened 2023-06-03
Adjust metadata writing sequence to avoid StreamerInfo missing

nkx111:
nkx111 Large: 3353 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/nkx111-streamerinfo) [](https://github.com/rest-for-physics/framework/commits/nkx111-streamerinfo)

This is to fix issue #358. Changed the sequence of metadata writing after process has finished. Previously: process metadata -- merge with process files -- TRestRun metadata. Now: merge with process files -- TRestRun metadata -- process metadata. Sometimes the merging operation will cause the later metadata writing operation overwriting the StreamerInfo of the previously written metadata classes. So we need to do the merging at the beginning.

This also fix a bug that the output file is not closed (because TRestRun is not deleted) by TRestManager. This might also be part of the cause of #358

### DIFF
--- a/source/framework/core/src/TRestManager.cxx
+++ b/source/framework/core/src/TRestManager.cxx
@@ -37,16 +37,28 @@ ClassImp(TRestManager);
 
 TRestManager::TRestManager() { Initialize(); }
 
-TRestManager::~TRestManager() {}
+TRestManager::~TRestManager() {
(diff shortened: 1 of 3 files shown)


nkx111:
Bad line endings! I will recreate the PR

PR #442, opened 2023-06-12
Do not exit when opening a non-rest file using `restRoot`

lobis:
lobis Ok: 90 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/lobis-trestrun-exit) [](https://github.com/rest-for-physics/framework/commits/lobis-trestrun-exit)

Currently opening a file from the terminal using `restRoot file.root` only works if the file has the full REST structure: `TRestRun`, `AnalysisTree` etc. Otherwise it returns an error.

It would make sense to also be able to open generic root files with `restRoot`. It should have the same effect as opening the files with root: loading objects into heap while also loading the REST libraries. This could be useful for example when opening a file containing only the readouts or some other metadata.

### DIFF
--- a/source/framework/core/src/TRestAnalysisPlot.cxx
+++ b/source/framework/core/src/TRestAnalysisPlot.cxx
@@ -82,7 +82,6 @@ void TRestAnalysisPlot::InitFromConfigFile() {
         // exit(1);
     }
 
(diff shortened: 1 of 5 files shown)


jgalan:
I just merged this PR into #441

PR #443, opened 2023-06-13
TRestDataSet fTimeCorrection data member added

Fonsi7:


source/framework/core/src/TRestDataSet.cxx
```diff
@@ -917,3 +926,66 @@ void TRestDataSet::Import(std::vector<std::string> fileNames) {
...
+                        corrected_time += t2 - t1;
+                        events_below_threshold = 0;
+                    }
+                }
+            }
+            if ((ev == r.GetEntries() - max_events_below_threshold) &&
+                (events_below_threshold > max_events_below_threshold)) {
+                last_bad_event = r.GetEntries() - 1;
+                r.GetEntry(first_bad_event);
+                t1 = r.GetAnalysisTree()->GetObservableValue<double>("rateAna_SecondsFromStart");
```
jgalan, line 982: Here you should use `fColumnTimeCorrection` instead of hard coding `rateAna_SecondsFromStart`.

jgalan, line 982: Here you should use `fColumnTimeCorrection` instead of hard coding `rateAna_SecondsFromStart`.

jgalan, line 982: This observable might not be in the analysis tree if the process `rateAna` was not used in the data processing., You should use `timeStamp`.

jgalan:
The `dataset.rml` is not present, to check that it is coherent with the code.

jgalan:
Inside `pipeline/datasets/` there should be the validation macro and the rml.

cmargalejo (approved):

PR #447, opened 2023-06-15
Simplifying opening of TRestRun and TRestDataSet files

juanangp:
juanangp Ok: 68 [](https://github.com/rest-for-physics/framework/commits/restroot_update)

Simplifiying `restRoot` behaviour for opening `TRestRun` and `TRestDataSet`
- Reimplementing TRestAnalysisTree, TRestEvent and metadata objects in a dedicated macro `REST_LoadInputFiles.C`
- Droping support for multiple files, now only one `TRestRun` and/or `TRestDataSet` can be opened (the first one)
- If the file is not a `TRestRun` or a `TRestDataSet` it just use default `root` behaviour using `TRint`
- Now the flag to load macros is just  `--m` instead of `--m 1` since the latest is misleading and error prone.

### DIFF
--- a/cmake/thisREST.cmake
+++ b/cmake/thisREST.cmake
@@ -136,7 +136,7 @@ export LIBRARY_PATH=\\\$REST_PATH/lib:\\\$LIBRARY_PATH
 export PYTHONPATH=${PYTHON_BINDINGS_INSTALL_DIR}:\\\$PYTHONPATH
 
 alias restRoot=\\\"restRoot -l\\\"
-alias restRootMacros=\\\"restRoot -l --m 1\\\"
+alias restRootMacros=\\\"restRoot -l --m\\\"
 
 if [ \\\$(rest-config --flags | grep \\\"REST_WELCOME=ON\\\") ]; then
     rest-config --welcome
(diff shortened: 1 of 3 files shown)


jgalan:
But `analysisTree`, `eventTree` and metadata objects have been removed from the automatic instances  ... I find them handy. 

It is not better to simplify it in #441?

juanangp:
> But `analysisTree`, `eventTree` and metadata objects have been removed from the automatic instances ... I find them handy.
> 
> I agree you simplify things, but without removing functionalities.
> 
> It is not better to simplify it in #441?

Ok, now the opening of input files is done using a macro, I didn't implement metadata  yet

juanangp:
> > But `analysisTree`, `eventTree` and metadata objects have been removed from the automatic instances ... I find them handy.
> > I agree you simplify things, but without removing functionalities.
> > It is not better to simplify it in #441?
> 
> Ok, now the opening of input files is done using a macro, I didn't implement metadata yet

Now metadata objects are also implemented, note that this PR is pointing to #441 instead of master. Please, do this merge before merging #441 in master.

jgalan (approved):

jgalan:
Perhaps this PR can be merged directly to master? Don't see why we should merge to #441

PR #392, opened 2023-03-27
New metadata classes to perform calibration and odds analysis over a TRestDataSet

juanangp:
juanangp Large: 738 [](https://gitlab.cern.ch/rest-for-physics/framework/-/commits/calibOdds) [](https://github.com/rest-for-physics/framework/commits/calibOdds)

New metadata clasess have been added under framework:
- `TRestCalibration` is meant to perform the calibration of a `TRestDataSet`, the  calibration is performed over different peaks provided in the config file. If an output file name is provided a new dataset will be generated with the corresponding metadata and calibration constants, while adding a new observable, `calib_energy`. In addition, the spectrum and the fit results will be stored in the output file.
- `TRestOdds`  performs the log odds of the different observables given in the config file and for a particular dataSet. To perform the log odds first the probability density funcion (PDF) is obtained for a set of observables in the desired range and used to compute the log odds. New observables are created in the output dataSet `odds_obsName` and the addition of all of them in `odds_total`. If an input odds file is provided, the different PDFs are retrieved from the input file.

PR #448, opened 2023-06-15
Implementing root TMVA for high level analysis

juanangp:
Implementing `TMVA` methods for high level analysis, please check  https://root.cern.ch/download/doc/tmva/TMVAUsersGuide.pdf for further details on `TMVA`.
Summary of changes:
- New class `TRestDataSetTMVA` to evaluate different  TMVA methods on `TRestDatasets` note that a signal and a background dataSet must be provided together with different `TMVA` methods.  This class evaluate all the methods provided and generates a root file and a folder with the results.
- New class `TRestDataSetTMVAClassification` to performs the classification of a given `TRestDataSet` using as input the results of the TMVA evaluation methods generated using `TRestDataSetTMVA`. An output dataset is generated by
/// definining a new observable with the TMVA method e.g. BDT_score.
- Added example `tmva.rml`
- Minor updates on `TRestRun` and `TRestDataSet`


jgalan:
I don't think we should name all the classes using `TRestDataSet` I think we should assume these classes will use internally `TRestDataSet` as something natural.

I would only use the `DataSet` inside the class name when it is clear that we need to differentiate.

For example, inside the sensitivity classes, I plan to have `TRestComponent` as a base class for `TRestDataSetComponent` and `TRestFormulaComponent`.

But it seems to me that `TRestDataSetXYZ` is over used.

I think discrimination methods usage should be unified by an abstract class `TRestDiscrimination`. Then TMVA method would be named `TRestTMVADiscrimination` that uses a `TRestDataSet` as implemented in `TRestDiscrimination`.

juanangp:
> I don't think we should name all the classes using `TRestDataSet` I think we should assume these classes will use internally `TRestDataSet` as something natural.
> 
Well, I just followed the suggestions done here https://github.com/rest-for-physics/framework/pull/392#issuecomment-1485726154
On the other hand, I think we should differenciate between discrimination and data processing, the new TMVA classes are just used to define new observables using TMVA, but the event discrimination is not performed there. I think the TMVA or eithed the LogOdds clasess fits under analysis folder, but we can implement more generic methods for event discrimintation in a different folder.

I don't know how to proceed renaming these classes because is not clear to me the best way to add them to the repository.

jgalan:
> > I don't think we should name all the classes using `TRestDataSet` I think we should assume these classes will use internally `TRestDataSet` as something natural.
> 
> Well, I just followed the suggestions done here [#392 (comment)](https://github.com/rest-for-physics/framework/pull/392#issuecomment-1485726154) On the other hand, I think we should differenciate between discrimination and data processing, the new TMVA classes are just used to define new observables using TMVA, but the event discrimination is not performed there. I think the TMVA or eithed the LogOdds clasess fits under analysis folder, but we can implement more generic methods for event discrimintation in a different folder.
> 
> I don't know how to proceed renaming these classes because is not clear to me the best way to add them to the repository.

Ok, perhaps we should discuss this online

Issue #452, opened 2023-06-20
Recovering event pointer when using `restRoot file.root`


jgalan:
Before #441 we could access the event pointer stored inside the ROOT file. I.e. `TRestGeant4Event`, `TRestRawSignalEvent`, ... depending on which event was available.

Now we recover a pointer to the event tree. 

Could we recover back the old behaviour? For me it was interesting to have ready access to the event as `ev0`.

Also, why not use the pointer instance name as `md_mdName`? instead of `metadata0["restG4 run"]` which we get right now?

I also think the `0` could disappear now from instance names.

<img width="838" alt="Screenshot 2023-06-20 at 18 23 38" src="https://github.com/rest-for-physics/framework/assets/12447509/162d67d7-a025-4b80-ad6b-8213258a604f">

juanangp:
> Could we recover back the old behaviour? For me it was interesting to have ready access to the event as `ev0`.

Now the input `TRestEvent` is accessible via `ev`
> Also, why not use the pointer instance name as `md_mdName`? instead of `metadata0["restG4 run"]` which we get right now?

Now we have a map of `TRestMetadata` pointers, that's why we have the `metadata["metaName"]`, it is not possible to define a variable with a particular name pattern. This was done previously with `gROOT->ProcessLine(...);` that I think we should try to avoid.

> I also think the `0` could disappear now from instance names.

Done

jgalan:
However, the interpreter seems to do not like it. See issue #450

juanangp:
> However, the interpreter seems to do not like it. See issue #450

Weel, it is not possible to attach an `std::map<std::string, TRestMetadata*>` as  `TRestMetadata*` so the root interpreter gets confused. I have added a solution on #453 by using `gROOT->ProcessLine`. Please close the other issue...

PR #453, opened 2023-06-20
Adding TRestEvent ptr while opening input file

juanangp:
juanangp Ok: 62 [](https://github.com/rest-for-physics/framework/commits/issue452)

Addressing issue #452 
- Adding `TRestEvent` pointer while opening `TRestRun` input file
- Removal of the appending `0` for the global variable names
- Now metadata and input event are added using `gROOT->ProcessLine`

### DIFF
--- a/macros/REST_OpenInputFile.C
+++ b/macros/REST_OpenInputFile.C
@@ -1,9 +1,7 @@
 TRestRun* run = nullptr;
 TRestAnalysisTree* ana_tree = nullptr;
 TTree* ev_tree = nullptr;
-TRestEvent* ev = nullptr;
 TRestDataSet* dSet = nullptr;
-std::map<std::string, TRestMetadata*> metadata;
 
 void REST_OpenInputFile(const std::string& fileName) {
     if (TRestTools::isRunFile(fileName)) {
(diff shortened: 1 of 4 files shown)


macros/REST_OpenInputFile.C
```diff
@@ -1,40 +1,44 @@
...
-        printf("\nAttaching TRestAnalysisTree as ana_tree0...\n");
-        ev_tree0 = run0->GetEventTree();
-        printf("\nAttaching event tree as ev_tree0...\n");
+        run = new TRestRun(fileName);
+        printf("\nAttaching TRestRun %s as run...\n", fileName.c_str());
+        ana_tree = run->GetAnalysisTree();
+        printf("\nAttaching TRestAnalysisTree as ana_tree...\n");
+        ev_tree = run->GetEventTree();
+        printf("\nAttaching event tree as ev_tree...\n");
+        ev = run->GetInputEvent();
```
jgalan, line 17: In the previous `restRoot` version we were creating an instance of the specific event type. That would allow us to access specific methods. Of course, we will still have access to `PrintEvent`, `DrawEvent`, and other common methods.

```
                string eventType = runTmp->GetInputEvent()->ClassName();

                printf("Attaching event %s as ev%i...\n", eventType.c_str(), nFile);
                string evcmd = Form("%s* ev%i = (%s*)%s;", eventType.c_str(), nFile, eventType.c_str(),
                                    (PTR_ADDR_PREFIX + ToString(runTmp->GetInputEvent())).c_str());
```

But just wondering if there is a solution for that.

juanangp: I am afraid there is no solution for that, before it was possible because it was using `gROOT->ProcessLine(...)` but I think we shouldn't be using that. Alternatively, you can do in the root terminal `auto rawEvent = static_cast<TRestRawSignalEvent *> (ev)`

jgalan (approved):

DavidDiezIb (approved):

PR #458, opened 2023-07-10
Adding documentation to CheckRunFileList macro

jgalan:
jgalan Ok: 6 [](https://github.com/rest-for-physics/framework/commits/jgalan_macro_doc)

Just adding documentation to `CheckRunFileList` macro.

This macro can be used to identify files that were not properly closed, e.g.: after launching the processing to condor cluster.

```
restManager CheckRunFileList "/path/file*Pattern.root"
```

The documentation will be printed when using `REST_PrintMacros.C`.

### DIFF
--- a/macros/REST_CheckRunFileList.C
+++ b/macros/REST_CheckRunFileList.C
@@ -11,11 +11,14 @@ using namespace std;
 
 //*******************************************************************************************************
 //***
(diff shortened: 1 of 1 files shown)


jgalan:
Any ideas to improve the macro are welcome. I thought perhaps it might be interesting to return the list of filenames that cannot be open properly in a `std::vector <std::string>`.

PR #468, opened 2023-08-28
Add VETO hits type

lobis:
Add new hit type `VETO`


DavidDiezIb (approved):

jgalan:
Perhaps it would be good to say why the new type is required. Where it is used?

PR #475, opened 2023-09-18
Remove PandaX from validation pipeline

lobis:
lobis Ok: 93 [](https://github.com/rest-for-physics/framework/commits/remove-pandaX-validation) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=rest-for-physics&utm_campaign=badge_info)

In some offline discussions we have raised the issue that the current pandaX validation usually leads to problems. When you modify (even by a small change) one of the core funcitonalities of the framework (e.g. readouts) the pandaX validations fail.

Usually this failure is due to the validation files needing to be regenerated again. The problem is that pandaX files are not really well understood by any of the current maintainers of the framework and (IMHO) this leads to too much time spent working around this validation.

I propose to remove this files from the validation unless we find a volunteer to keep them up to date (or create documentation that explains this process in an easy way).

The pandaX validation should be replaced by a more detailed validation for some of the current experiments using REST-for-Physics, such as IAXO or TREX-DM (this to be done in a different PR).

Let me know your thoughts.


jgalan:
It makes sense for me, I think @nkx111 agreed with this, but just let him reply here

nkx111:
I propose to keep two simple tests from the pipeline: 1. running `restG4` with `Xe136bb0n.rml`, and 2. running `restManager` with `P3AutoChain.rml` to process graw data file. For other tests of file validation it is indeed troublesome and it is OK to remove.

nkx111 (approved):

juanangp (approved):

PR #485, opened 2023-09-26
Dictionary generation not working on batch jobs (condor)

lobis:
lobis Ok: 61 [](https://github.com/rest-for-physics/framework/commits/481-error-running-restmanager-using-condor) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=rest-for-physics&utm_campaign=badge_info)

After this PR, the environment varible `REST_HOME` will be used to define the "rest home directory" (currently `$HOME/.rest`. If `REST_HOME` does not exist, it will default to the old behaviour.

This is because using home directory can lead to problems in some cases, expecially when running batch jobs, the gdml geometry is replicated multiple times for each job, which causes the disk to fill up quite fast and leads to errors.

With this feature the user could use the `/tmp` directory (for example) to store these files.


jgalan (approved):
I approve, it would be interesting that `thisREST.sh` should define a default `REST_HOME`?

nkx111 (approved):

lobis:
> I approve, it would be interesting that `thisREST.sh` defines a default `REST_HOME`?

Done in https://github.com/rest-for-physics/framework/pull/485/commits/9f916cfe44e707abdd2c22a000eadc40dd5b2977. With this change users should be aware that this env variable should be overrided **after** loading `thisREST.sh` (before this change it didn't matter).

Issue #486, opened 2023-10-05
Merging datasets is not properly handled


juanangp:
I found an issue merging datasets using `TRestDataSet::Import` function implemented here:
https://github.com/rest-for-physics/framework/blob/9c2fae713e932c11c23efc78e332bdac2e3dad5e/source/framework/core/src/TRestDataSet.cxx#L952

Although the trees are properly merged the metadata has issues, for instance while merging two datasets the duration is not properly retreived.

I guess we should implement a TRestDataSet merging of the metadata to solve this issue

PR #487, opened 2023-10-05
Dataset merge issue

juanangp:
juanangp Ok: 65 [](https://github.com/rest-for-physics/framework/commits/dSMergeIssue) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=rest-for-physics&utm_campaign=badge_info)

Implementation of merging of metadata while importing more than one dataset, in addition any file inside the list which is not a valid dataSet will be skipped.

Address issue #486

### DIFF
--- a/source/framework/core/src/TRestDataSet.cxx
+++ b/source/framework/core/src/TRestDataSet.cxx
@@ -894,7 +894,6 @@ TRestDataSet& TRestDataSet::operator=(TRestDataSet& dS) {
     fFilterEqualsTo = dS.GetFilterEqualsTo();
     fQuantity = dS.GetQuantity();
     fTotalDuration = dS.GetTotalTimeInSeconds();
-    fFileSelection = dS.GetFileSelection();
     fCut = dS.GetCut();
 
     return *this;
--- a/source/framework/core/src/TRestDataSet.cxx
+++ b/source/framework/core/src/TRestDataSet.cxx
@@ -912,19 +912,11 @@ Bool_t TRestDataSet::Merge(const TRestDataSet& dS) {
         }
     }
 
-    if (REST_StringHelper::StringToTimeStamp(fFilterStartTime) >
-        REST_StringHelper::StringToTimeStamp(dS.GetFilterStartTime()))
-        fFilterStartTime = dS.GetFilterStartTime();
-    if (REST_StringHelper::StringToTimeStamp(fFilterEndTime) <
-        REST_StringHelper::StringToTimeStamp(dS.GetFilterEndTime()))
-        fFilterEndTime = dS.GetFilterEndTime();
     if (fStartTime > dS.GetStartTime()) fStartTime = dS.GetStartTime();
     if (fEndTime < dS.GetEndTime()) fEndTime = dS.GetEndTime();
 
     auto fileSelection = dS.GetFileSelection();
     fFileSelection.insert(fFileSelection.end(), fileSelection.begin(), fileSelection.end());
-
-    fProcessObservablesList = dS.GetProcessObservablesList();
 
     fTotalDuration += dS.GetTotalTimeInSeconds();
 
--- a/source/framework/core/src/TRestDataSet.cxx
+++ b/source/framework/core/src/TRestDataSet.cxx
@@ -904,20 +904,22 @@ TRestDataSet& TRestDataSet::operator=(TRestDataSet& dS) {
 /// metadata in current dataSet
 ///
 Bool_t TRestDataSet::Merge(TRestDataSet& dS) {
-
     auto obsNames = GetObservablesList();
-      for (const auto& obs : fObservablesList) {
+    for (const auto& obs : fObservablesList) {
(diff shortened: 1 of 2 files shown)


source/framework/core/src/TRestDataSet.cxx
```diff
@@ -894,12 +894,43 @@ TRestDataSet& TRestDataSet::operator=(TRestDataSet& dS) {
...
+    for (const auto& obs : fObservablesList) {
+        if (std::find(obsNames.begin(), obsNames.end(), obs) != obsNames.end()) {
+            RESTError << "Cannot merge dataSets with different observable list " << RESTendl;
+            return false;
+        }
+    }
+
+    if (REST_StringHelper::StringToTimeStamp(fFilterStartTime) >
+        REST_StringHelper::StringToTimeStamp(dS.GetFilterStartTime()))
+        fFilterStartTime = dS.GetFilterStartTime();
```
restforphysics, line 917: This is a bit tricky because a generated dataset 1 with a given startTime filter, and dataset 2 with filter 2, then to be rigorous I need to know the startTime filter of both.

So that I know that the dataset 1 was produced with a given filter, and the other was produced with a different time filter. Idem for other filters.

Perhaps a new class `TRestDataSetMerged` could help. We keep the filters only for `TRestDataSet` while in `TRestDataSetMerged` we define as metadata members the dataset filenames used for the merge?

And perhaps introduce a hash-id to identify/guarantee the TRestDataSet used to create the merged dataset?

jgalan: Ok that previous post was me

source/framework/core/src/TRestDataSet.cxx
```diff
@@ -894,12 +894,43 @@ TRestDataSet& TRestDataSet::operator=(TRestDataSet& dS) {
...
     fCut = dS.GetCut();
 
     return *this;
 }
 
+///////////////////////////////////////////////
+/// \brief This function merge different TRestDataSet
+/// metadata in current dataSet
+///
+Bool_t TRestDataSet::Merge(TRestDataSet& dS) {
```
lobis, line 906:
```suggestion
Bool_t TRestDataSet::Merge(const TRestDataSet& dS) {
```

We don't want a mutable reference here right? The input should be read-only (if I understood correctly).

juanangp: Indeed, should be const

Done

source/framework/core/src/TRestDataSet.cxx
juanangp: I just removed the timeStamp filters, I think that to make a new class TRestDataSetMerged is an overkill.

However, I don't know what is the best strategy here because apart of the filters there is plenty of metadata info which is not saved when we merge several dataSets. Note that before just only the metadata info from the first file was saved.

jgalan: Actually, for me a dataset does not need to contain all the metadata information that is available to the run files. The dataset must guarantee that we can identify the run files used to generate the dataset. I guess, having the selected files vector is already a good starting point. Then we need to know if any filters have been applied for event selection. But the final user working with the dataset needs to know also other conditions applied to generate the dataset, e.g. the detector pressure, the data taking days, the voltage on the mesh, etc, etc. Then, if those filters were used to generate the datasets, it is relevant for the end-user.

The metadata information transferred to the dataset in the form of relevant quantities is to provide the dataset with physical meaning, and allow to have all the analysis required information into one single entity.

Indeed, if we added new columns to the dataset, we need to add also that metadata information, because if I generate a new column using `<addColumn` then I need to now the formula used to calculate that column. Not sure if we are doing this right now.

jgalan: Ok, I see now that the columns added should be in the form `fColumnNameExpressions`. Just last time I couldn't find them in my dataset.

jgalan: > Note that before just only the metadata info from the first file was saved.

In my opinion this should never happen, as I already stated many times I am not agree on merging `runs`. Merging `datasets` is different issue, but `runs` should be considered as `read-only` entities, that can be used later on to generate a data compilation or dataset.

juanangp: > > Note that before just only the metadata info from the first file was saved.
> 
> In my opinion this should never happen, as I already stated many times I am not agree on merging `runs`. Merging `datasets` is different issue, but `runs` should be considered as `read-only` (or end-product of event data processing) entities, that can be used later on to generate a data compilation or dataset.

I am talking about `datasets,` not `TRestRuns` the issue is that if the metadata info of the dataset is not properly handled the results are wrong when I compute e.g. the rate. In this case the duration of the dataset has to be added, but other metadata info can be also wrong or misleading. I don't know what is the proper way to handle metadata info of the dataset while merging, but this issue was present before this PR.

jgalan (approved):

lobis (approved):

PR #355, opened 2022-12-19
Implementation of experimental sensitivity classes

jgalan:
jgalan Large: 3739 [](https://github.com/rest-for-physics/framework/commits/jgalan_sensitivity) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=rest-for-physics&utm_campaign=badge_info)

- [x] Implement TRestComponent
- [x] Implement TRestComponentDataSet
- [x] Implement TRestComponentFormula
- [x] Implement TRestExperiment
- [x] Implement TRestExperimentList 
- [x] Implement TRestResponse
- [x] Implement TRestSensitivity

Other add-onds that were required during the development:

- Method `REST_StringHelper::RemoveDelimiters` was implemented in order to remove any of [{()}] from an input string.
- Method `REST_StringHelper::IntegerToBinary` is used to return a `std::vector` containing 0s and 1s, that reproduce the integer number in binary format. This is used on the N-dimensional interpolation, in order to access the interpolation points, reducing the integer number in an iterative way to a base of the N-dimensional space.
- Method `TRestTools::CanvasDivisions(int n)` to help finding a good `NxM` ratio to host `n` plots inside a canvas.
- `TRestTools` now supports also `std::string` tables.

PR #506, opened 2024-02-11
CMakeLists. Adding GSL libraries

jgalan:


nkx111:
How is GSL library used by REST? I think we need to set the default option as OFF in CMakeLists.txt

jgalan:
> How is GSL library used by REST? I think we need to set the default option as OFF in CMakeLists.txt

I thought GSL libraries are quite standard libraries found on any system.

Right now we are using them only here:

https://github.com/rest-for-physics/axionlib/blob/master/src/TRestAxionField.cxx

Perhaps we could simply disable them as soon as RESTLIB_AXION is not ON.

jgalan:
@nkx111 I have implemented the condition for GSL libraries being linked here:

https://github.com/rest-for-physics/framework/pull/355/commits/a1ff7cdbbcf7d6d2545307d67525fbde0d09a2d9


But I will close this PR since this is already implemented at PR #355

PR #514, opened 2024-04-10
Adding possibility to add a list of files to TRestDataSet 

juanangp:
Added the posibility to add a file list instead of a file pattern in TRestDataSet.

Note that if the file list is empty it will just use the file pattern provided, however the file pattern will be ignored if the file list is not empty.

### DIFF
--- a/source/framework/core/inc/TRestDataSet.h
+++ b/source/framework/core/inc/TRestDataSet.h
@@ -213,6 +213,6 @@ class TRestDataSet : public TRestMetadata {
     TRestDataSet(const char* cfgFileName, const std::string& name = "");
     ~TRestDataSet();
 
-    ClassDefOverride(TRestDataSet, 7);
+    ClassDefOverride(TRestDataSet, 8);
 };
 #endif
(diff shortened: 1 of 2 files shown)


AlvaroEzq:
Maybe it could be generalized to a list of file patterns? I have found myself in this situation where one file pattern isn't enough to get all the files I need. 

I found out that TRestTools::GetFilesMatchingPattern() accepts several file patterns by joining them with "\n" (for example "/data/R00001\_\*\_tags.root\n/data/R00002\_\*\_tags.root") and I have using this workaround but is quite annoying. This will be a more readable solution.

jgalan (approved):
Perhaps when we call `SetFileList` and the list is not empty, we should force `fFilesMatchingPattern = "fileList"` so that when we explore the metadata we know the file list was initialized by other means other than a filename pattern.

juanangp:
> Perhaps when we call `SetFileList` and the list is not empty, we should force `fFilesMatchingPattern = "fileList"` so that when we explore the metadata we know the file list was initialized by other means other than a filename pattern.
> 
> Or, following @AlvaroEzq proposal we may promote `std::string fFilesMatchingPattern` to a `std::vector <std::string> fFilesMatchingPattern;` which can be either a pattern name list or just a file list.

Ok, now `fFilePattern` is a vector of strings that can be set using `SetFilePattern`, it should be also compatible with previous implementation.

PR #526, opened 2024-06-14
Creating new TRestRadialStrippedMask

jgalan:
jgalan Large: 303 [](https://github.com/rest-for-physics/framework/commits/jgalan_radial_good) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=rest-for-physics&utm_campaign=badge_info)

This PR adds a new type of radial mask that will be required by new custom optics  https://github.com/rest-for-physics/axionlib-data/pull/11

radialstrippedmask

Screenshot 2024-06-14 at 15 59 07


juanangp (approved):

source/framework/masks/src/TRestRadialStrippedMask.cxx
```diff
@@ -0,0 +1,213 @@
...
+///
+/// The particle will be counter-rotated to emulate the mask rotation
+/// using the method TRestPatternMask::ApplyCommonMaskTransformation
+///
+Int_t TRestRadialStrippedMask::GetRegion(Double_t& x, Double_t& y) {
+    if (TRestPatternMask::GetRegion(x, y)) return 0;
+
+    TVector2 point(x, y);
+    Double_t phi = point.Phi();
+
```
jovoy, line 168: So one thing that we still need for this is the initialRadius like in the TRestSpiderMask. Does this mean that X-rays hitting inside the initialRadius are stopped? And then if you use internalRegionRadius they pass in that region? If I understood that correctly we would also need internalRegionRadius because some of the Spacers only start at the halfway mark of the radii.

jgalan: Ok, I have updated accordingly to follow the same logic as TRestSpiderMask. I updated the image on the documentation. Watch out the `masks.rml` file.

jovoy (approved):

PR #528, opened 2024-06-24
TRestEventSelectionProcess upgrade

AlvaroEzq:
AlvaroEzq Ok: 18 [](https://github.com/rest-for-physics/framework/commits/aezquerro_eventSel) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=rest-for-physics&utm_campaign=badge_info)

Adding new possibility of using the TRestEventSelectionProcess without an external fileWithIDs by evaluating the conditions on the TRestAnalysisTree of the processing file itself.

### DIFF
--- a/source/framework/analysis/inc/TRestEventSelectionProcess.h
+++ b/source/framework/analysis/inc/TRestEventSelectionProcess.h
@@ -24,17 +24,16 @@
 #define RestProcess_TRestEventSelectionProcess
 
 #include <TH1D.h>
+#include <TRestEventProcess.h>
 
 #include <iostream>
-
-#include "TRestEventProcess.h"
 
 //! A template process to serve as an example to create new TRestRawSignalEventProcess
 class TRestEventSelectionProcess : public TRestEventProcess {
    private:
     TRestEvent* fEvent;  //!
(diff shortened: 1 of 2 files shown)


DavidDiezIb (approved):

DavidDiezIb:
I guess this was to include the selection process in the processing chain, right? Before it was only possible from existing files, not in the middle of the processing chain, am I correct?

AlvaroEzq:
> I guess this was to include the selection process in the processing chain, right? Before it was only possible from existing files, not in the middle of the processing chain, am I correct?

Yes!

lobis:
@AlvaroEzq I approve the PR but made some minor changes which shouldn't modify the behaviour:

- `std::string` are already initialized as `""`, no need to explicitly do it.
- Don't use `new` / `delete` if it can be avoided, in this case you can just create a regular (stack allocated) `TRestRun` which will be deleted when existing the scope.
- Whenever you can try to use `{}` for all `if` statements and such, in my experience not using them for single-line expressions usually leads to mistakes down the road.

lobis (approved):

PR #532, opened 2024-07-13
Adding LCG to thisREST.sh if LCG environment is identified

jgalan:
jgalan Ok: 7 [](https://github.com/rest-for-physics/framework/commits/jgalan_lcg) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=rest-for-physics&utm_campaign=badge_info)

The following PR adds some code inside `thisREST.cmake` to identify if we are using LCG environment. It will do so in case `LCG_VERSION` is found. And if it is found it will use the path defined by `C_INCLUDE_PATH` to load the environment inside `thisREST.sh`.

- Some issues inside `cmake_uninstall.cmake.in` have also corrected.

### DIFF
--- a/cmake/thisREST.cmake
+++ b/cmake/thisREST.cmake
@@ -5,7 +5,7 @@ set(lcg_found FALSE)
 
 # Check if the CMAKE_COMMAND contains both "cvmfs", "lcg" and "x86_64-el9-gcc11"
 string(FIND "${CMAKE_COMMAND}" "cvmfs" cvmfs_pos)
-string(FIND "${CMAKE_COMMAND}" "LCG_104" lcg_pos)
+string(FIND "${CMAKE_COMMAND}" "lcg" lcg_pos)
 string(FIND "${CMAKE_COMMAND}" "x86_64-el9-gcc11" gcc_pos)
 
 if (lcg_pos GREATER -1
@@ -17,7 +17,7 @@ endif ()
 set(loadLCG "")
 if (lcg_found)
     set(loadLCG
-        "\# REST_LCG was enabled. We load LCG_104 environment.\nsource /cvmfs/sft.cern.ch/lcg/views/LCG_104/x86_64-el9-gcc11-opt/setup.sh\n\n"
+        "\# We load LCG_104 environment.\nsource /cvmfs/sft.cern.ch/lcg/views/LCG_104/x86_64-el9-gcc11-opt/setup.sh\n\n"
     )
 endif ()
 
--- a/cmake/thisREST.cmake
+++ b/cmake/thisREST.cmake
@@ -5,7 +5,7 @@ set(lcg_found FALSE)
 
 # Check if the CMAKE_COMMAND contains both "cvmfs", "lcg" and "x86_64-el9-gcc11"
 string(FIND "${CMAKE_COMMAND}" "cvmfs" cvmfs_pos)
-string(FIND "${CMAKE_COMMAND}" "lcg" lcg_pos)
+string(FIND "${CMAKE_COMMAND}" "LCG_104" lcg_pos)
 string(FIND "${CMAKE_COMMAND}" "x86_64-el9-gcc11" gcc_pos)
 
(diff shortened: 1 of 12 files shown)


cmake/thisREST.cmake
```diff
@@ -1,5 +1,17 @@
 # Write thisREST.[c]sh to INSTALL directory
 
+##### Checking if we are using CVMFS LCG GCC11 environment
+set(lcg_found FALSE)
+
+# Check if the CMAKE_COMMAND contains both "cvmfs", "lcg" and "x86_64-el9-gcc11"
```
LouisHelary, line 6: A test like this one (especially the one: x86_64-el9-gcc11 ) will make it more difficult to work if we have to change the release or the OS it is compiled against for whatever reason. Can we try to change it to a test based on the kernel number for instance?

LouisHelary (commented):
Besides the comment I have left on the OS test, I would also try not to be too specific about the LCG release we currently use (that might change in the future). But otherwise the code looks good.

cmake/thisREST.cmake
jgalan: This is what I could think of. Not sure about the kernel thing. How the kernel version relates to LCG_104?

jgalan:
> Besides the comment I have left on the OS test, I would also try not to be too specific about the LCG release we currently use (that might change in the future). But otherwise the code looks good.

The point here is to identify if the user has loaded LCG_104. I don't really know how to do this, perhaps there is a good way? I just check that the `cmake` command path is coming from `/cvmfs/`, `lcg` and `x86_64-el9-gcc11`. If we are able to identify which script was sourced in the environment, then probably we wold find a more generic and clean way to do this.

jgalan:
Apparently there is a LGC_VERSION defined as system variable: 

```
[jgalan@naf-iaxo21 rest-framework]$ env | grep LCG_VERSION
LCG_VERSION=104
```

LouisHelary:
> jgalan Ok: 7 [](https://github.com/rest-for-physics/framework/commits/jgalan_lcg) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=rest-for-physics&utm_campaign=badge_info)
> 
> The following PR adds some code inside `thisREST.cmake` to identify if we are using `/cvmfs/sft.cern.ch/lcg/views/LCG_104/x86_64-el9-gcc11-opt/setup.sh`. In case it is found, `thisREST.sh` will load first that script.
> 
> I have observed the following problem when using /cvmfs/.../lcg../ installation.
> Screenshot 2024-07-13 at 20 50 11

I observe the same error, but only if I dont compile a release. If I use the script  ./install_rest_release_LCG.sh v2.4.3 instead I can't reproduce the problem. Do you have older compilation logs available?

Install the project...
-- Install configuration: ""
CMake Warning at cmake_uninstall.cmake:2 (message):
  Cannot find install manifest:
  /tmp/cronjob-rest-release-v2.4.3/build/install_manifest.txt


-- Installing: /nfs/dust/iaxo/user/lhelary/LCG_rest/test/install/lib/libRestFramework.so

jgalan:
Hi @lobis , there are issues with the certificate when downloading from sultan. Certificate validation may be set as a ROOT configuration parameter. Perhaps it can be changed in the docker image?

https://root-forum.cern.ch/t/error-in-davixopen-can-not-open-file-with-davix/19355

lobis:
> Hi @lobis , there are issues with the certificate when downloading from sultan. Certificate validation may be set as a ROOT configuration parameter. Perhaps it can be changed in the docker image?
> 
> https://root-forum.cern.ch/t/error-in-davixopen-can-not-open-file-with-davix/19355
> 

I think this is just an issue with the web server, it shouldn't take long to fix (hopefully) on the server's end. We are aware of the issue.

jgalan:
Great, then I think this PR is ready to be merged.

LouisHelary (approved):

lobis (approved):

PR #558, opened 2026-01-12
Fix TREX-DM test

AlvaroEzq:
AlvaroEzq Ok: 10 [](https://github.com/rest-for-physics/framework/commits/aezq_fixCertificatesCI) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=rest-for-physics&utm_campaign=badge_info)

The new sultan certificate has to be imported manually into the docker image. So I created a clone of @lobis 's one and add to it the import of the current sultan CA certificate.

Even doing this, ROOT cannot download it (although wget yes) because ROOT uses Davix for remote HTTPS access, but Davix relies on the CA bundle that was compiled into the ROOT binary. In our environment, the Sultan institutional certificates are not included in that bundle, and there’s no runtime way to add them without rebuilding ROOT. As a result, attempts to open the file directly in ROOT fail with certificate verification errors. That's why I move the download of the files outside ROOT from the [classify.rml](pipeline/analysisPlot/classify.rml) file

EDIT: at the end the new image is not needed so I go back to lobis one. It is just needed to manually import the sultan certificates.

### DIFF
--- a/.github/workflows/validation.yml
+++ b/.github/workflows/validation.yml
@@ -227,14 +227,9 @@ jobs:
     name: "TREX-DM"
     runs-on: ubuntu-latest
     container:
-      image: ghcr.io/lobis/root-geant4-garfield:rest-for-physics
+      image: ghcr.io/alvaroezq/root-geant4-garfield:rest-for-physics-sultancertificate
     needs: [ framework-install ]
     steps:
-      - name: Fix TLS certificates
-        run: |
-          apt-get update
-          apt-get install -y --reinstall ca-certificates openssl
-          update-ca-certificates --fresh
       - uses: actions/checkout@v4
         with:
           repository: rest-for-physics/framework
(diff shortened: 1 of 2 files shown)


AlvaroEzq:
Perhaps is better idea to just add a job to download the sultan certificate and update-ca-certificates, so we can keep @lobis 's docker image as it was?

lobis:
At some point the repo https://github.com/rest-for-physics/rest-docker was created to use this image instead of the one in my personal repo, there is even a PR for this: https://github.com/rest-for-physics/framework/pull/365.

There was some issue preventing from merging this PR and it was left undone, but this might be a good opportunity to finish the job!

AlvaroEzq:
> At some point the repo https://github.com/rest-for-physics/rest-docker was created to use this image instead of the one in my personal repo, there is even a PR for this: #365.
>
> There was some issue preventing from merging this PR and it was left undone, but this might be a good opportunity to finish the job!

Nice! Thanks. I will try to see if I can sort that issue... :smile:


In the meantime, I would like to merge this PR.

AlvaroEzq:
> Perhaps is better idea to just add a job to download the sultan certificate and update-ca-certificates, so we can keep @lobis 's docker image as it was?

I went back to lobis image to keep the changes more simple.

mariajmz (approved):

juanangp (approved):

Issue #542, opened 2025-01-13
Huge memory leak when opening files made with REST<latest (?)


AlvaroEzq:
There is a memory leak when opening, with latest version of REST, a REST files made with version < 2.4.3 or <=2.4.2 (i am not sure). The memory leak is caused by opening the event tree with TRestDetectorHints events and just end up saturating the system memory.

AlvaroEzq:
Maybe is the same memory leak reported in this previous issue https://github.com/rest-for-physics/framework/issues/519

PR #563, opened 2026-04-17
Prevent hanging when opening old REST files with schema mismatches

cmargalejo:
cmargalejo Ok: 22 [](https://github.com/rest-for-physics/framework/commits/cris_fix_old_file_hang) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=rest-for-physics&utm_campaign=badge_info)  

This PR prevents `restRoot` from hanging when opening files produced with older REST versions.

  ## Context
When a REST file is produced, each class (e.g. `TRestDetectorSignalToHitsProcess`) is saved with a specific internal layout: which fields, in which order, how many bytes. If the class changes in a newer version, this layout no longer matches. ROOT tries to reconcile the old data with the new layout, but in some cases gets stuck, causing `restRoot` to hang at `GetEntry(0)`. This fix detects those errors and skips the event loading, printing a warning instead. The analysis tree (`ana_tree`) and metadata remain accessible. Files produced with compatible versions are (or at least should be) unaffected.

How it works
- ROOT provides a function called `SetErrorHandler()` that allows replacing the default error printing with a custom function.
- We use this to temporarily install our own handler (`SchemaErrorHandler`) that does two things. It prints the error as usual, and sets a flag if the error is related to schema mismatches.
- Before calling `GetEntry(0)` (the line that causes the hang), we check the flag. If schema errors were found, we skip `GetEntry(0)` and print a warning instead.
- The custom handler is only active while the file is being opened. After that, ROOT's default handler is restored

Tested with two files:
- v2.3.6 file (schema mismatch): no longer hangs, warning printed, `ana_tree` accessible
- v2.4.0 file (compatible): works normally with full event browsing

  Relates to #542

  @rest-for-physics/core_dev

### DIFF
--- a/macros/REST_OpenInputFile.C
+++ b/macros/REST_OpenInputFile.C
@@ -1,3 +1,16 @@
+bool gSchemaError = false;
+
+void SchemaErrorHandler(int level, Bool_t abort, const char* location, const char* msg) {
+    if (level >= kError) {
+        std::string loc(location);
+        if (loc.find("TBufferFile") != std::string::npos ||
+            loc.find("TStreamerInfo") != std::string::npos) {
+            gSchemaError = true;
+        }
+    }
+    DefaultErrorHandler(level, abort, location, msg);
+}
+
 TRestRun* run = nullptr;
 TRestAnalysisTree* ana_tree = nullptr;
 TTree* ev_tree = nullptr;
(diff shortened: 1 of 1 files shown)


cmargalejo:

juanangp (approved):

AlvaroEzq (approved):

PR #566, opened 2026-05-04
Add safe recovery for legacy detector signal schemas

cmargalejo:
cmargalejo Large: 4252 [](https://github.com/rest-for-physics/framework/commits/cris_legacy_detector_signal_branch) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=rest-for-physics&utm_campaign=badge_info)  

## Quick summary for reviewers:
This adds a standalone recovery tool for legacy detector-signal files (produced before the vector<double> schema). It's opt-in and doesn't change how REST normally reads or writes files, so it should be low-stress to review.

What it adds (the bulk):
- A restRoot --recover-legacy-signals INPUT [--output … | --in-place] command + two macros in macros/legacy/, plus tests. Fully self-contained — nothing calls into it during normal processing.

The only framework read-path change is defensive: when REST opens a file whose detector-signal branch uses the old layout, it now detects it, disables that branch, and warns, instead of attempting the read and hitting the >7 GB OOM. That's strictly safer than today's behaviour for anyone who opens an old file, and it's a no-op for normal files.

What it does not change:
- Normal read/write of current files — untouched.
- Anyone starting fresh after the next release never triggers any of this (they have no legacy files to recover).

How to review quickly: the recovery CLI/macros can be read in isolation; the only thing that affects existing behaviour is the warn-and-disable guard. Easy to revert — it's additive commits, so dropping it leaves the framework exactly as before.

Complements #567: #567 stops REST from producing these files going forward; #566 recovers the ones that already exist.

## *UPDATE* 2026-08-19:

This PR changed significantly from what it was initially supposed to do. Instead of just warning if unsupported files are detected, we now have a full blown data recovery feature in REST, which allows us to fix up old files. Essentially when we have an old file that does not yet contain `StreamerInfo` (and thus the schema change in `TRestRawSignal` from `vector<Float_t>` to `vector<Double_t>` of PR https://github.com/rest-for-physics/detectorlib/pull/109/ causes a faulty data read leading to an out-of-memory error), we can reinsert that information by rewriting the input file (technically writing a copy) to have that `StreamerInfo`, making it readable by current and future REST versions. PR #567 will then handle the case that we don't accidentally lose that information in the future.

The large increase in code is mainly because this PR has evolved from a warning-only change into a complete, tested recovery workflow for affected legacy files:

- The PR is now rebased onto master and fully decoupled from #564.
- `restRoot --recover-legacy-signals INPUT [--output OUTPUT | --in-place]` provides a safe one-command recovery path.
- Recovery validates the legacy schema and reconstructed data before replacing anything, preserves unrelated top-level trees, avoids overwriting files during races, and retains a backup for in-place recovery.
- The warning in `TRestRun` now detects unsafe explicit reads and points users to the recovery command.
- Most of the additional code consists of validation, error handling, regression tests, and user documentation rather than changes to normal event processing.

The implementation was tested both independently and together with #567, including against the original reproducer. The combined focused suite passes 46/46 tests.

### Original PR description

This PR is built on top of #564 and should be reviewed/merged after #564 (I keep it separate because maybe we don't want to change TRestRun.cxx even more just to print a warning).

It mitigates rest-for-physics/detectorlib#125, but it's not a real fix. It just avoids crashing.

Some old files contain `TRestDetectorSignalEvent` branches where `TRestDetectorSignal::fSignalTime` and `fSignalCharge`  were written as  `vector<float>`. Current detectorlib expects these members as `vector<double>` after detectorlib#109.

This PR detects that unsupported branch, prints a warning, and disables only that signal event branch.  Other events types remain readable.

## Update

This PR now also includes the solution, not just the warning:

- `macros/legacy/recoverLegacySignalData.C` (stage 1, plain `root`): extracts the legacy float signal data using replica classes matching the old on-disk layout exactly.
- `macros/legacy/REST_RebuildLegacySignalFile.C` (stage 2, `restRoot`): writes a fixed file with the signal branch rebuilt using the current classes, all other branches/metadata copied, and StreamerInfos stored this time. Optional in-place overwrite (keeps a `.bak`).
- The warning now prints these two commands.

~~Depends on #564 (base branch).~~

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -295,7 +295,15 @@ install(
 install(
     DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/macros
     DESTINATION .
-    COMPONENT install)
+    COMPONENT install
+    PATTERN "REST_RebuildLegacySignalFile.C" EXCLUDE)
+if (TARGET RestDetector)
+    install(
+        FILES
+            ${CMAKE_CURRENT_SOURCE_DIR}/macros/legacy/REST_RebuildLegacySignalFile.C
+        DESTINATION macros/legacy
+        COMPONENT install)
+endif ()
 install(
     DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/data
     DESTINATION .
--- a/doc/tutorials/List of REST Programs.md
+++ b/doc/tutorials/List of REST Programs.md
@@ -8,6 +8,8 @@ This is just an indexing list of existing programs included with REST compilatio
 
 **restManager** : The manager program of REST allowing to execute the REST event processes defined in TRestManager. Additional definitions allow to include additional metadata structures inside REST.
 
+**restRoot** : The ROOT interactive shell with REST libraries and helpers loaded. It also provides the safe one-command legacy detector-signal recovery workflow described in [Recovering legacy detector signal files](<Recovering legacy detector signal files.md>).
+
 **restPlots** : It uses the plot definitions given by RML configuration in a TRestAnalysisPlot section. It creates the plots from the variables at the TRestAnalysisTree and creates a PDF report and a ROOT file including the histograms created.
 
 ## Histogram and integration executables
--- a/doc/tutorials/Recovering legacy detector signal files.md
+++ b/doc/tutorials/Recovering legacy detector signal files.md
@@ -1,0 +1,112 @@
+# Recovering legacy detector signal files
+
+REST files written with detectorlib versions before the `TRestDetectorSignal`
+schema update stored signal time and charge arrays as `vector<float>`. Current
+detectorlib uses `vector<double>`. Some old files do not contain enough ROOT
+streamer information to convert that change safely; directly selecting the
+legacy detector-signal branch can otherwise cause a very large allocation,
+crash, or out-of-memory failure.
+
+The recovery command supports the known legacy signal schema versions 1, 2,
+and 3:
+
+```sh
+restRoot --recover-legacy-signals input.root
+```
+
+The default result is `input_Fixed.root`. The input is not modified. Choose a
+different new output path with:
+
+```sh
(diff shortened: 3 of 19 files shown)


Vindaar (approved):
With the improvements this PR is now useful to not just warn the user, but also allow reading old files.

Ideally a few people would read all the diff, but given the size of it that is not very realistic. However, please note that the majority of the new added lines are *NOT* part of the REST core. The majority just deal with the data recovery path that was added. Hence, it's not a new maintenance burden of 4k lines for REST itself (and well, the need for this is our own fault for not making sure we preserve `StreamerInfo` in the first place).

cmargalejo:
Hey @Vindaar, i have played with the recovery on a real file (R00236, 261 entries) and found a couple of things the tests didn't, so I pushed two commits:

1bec4385 — segfault fix. Real files have old process metadata whose classes were since renamed/removed (TRestRawZeroSupressionToRawProcess). ROOT builds a non-null emulated TClass for those, so the cl != nullptr guard passed and ReadObj() crashed on an abstract base in the chain. Now I also skip classes with no dictionary (!HasDictionary()) → they go through the normal skip-and-warn path. No fixture had an obsolete class, so nothing caught it.

9492996b — usability. Two things: (1) the CLI forced REQUIRE_COMPLETE=1, so it refused any file with an unreadable class — i.e. the exact files we want to fix. Made partial/skip-and-warn the default, with --require-complete to opt into strict (in-place still refuses partial on its own). (2) It rejected restRoot -l …, but thisREST.sh aliases restRoot to restRoot -l, so the documented command needed \restRoot. Now it accepts the leading -l.

Small note: the branch still pins detector at the old v3/float commit, so an as-pinned build fails with "detected version 3." Nothing to do in detectorlib (master's already v4) — just needs the pin ≥ v4 at release. I left the pointer alone and will mention it in the PR description.

Give it a look — happy to change anything.

PR #567, opened 2026-06-12
Centralize schema-safe writable ROOT I/O (preserve `StreamerInfo` on ROOT ops)

### LINKED ISSUE #568: Output files lose event-class StreamerInfos (UPDATE session on close wipes them)
## Summary

Every output file written by restManager is missing the StreamerInfos of the event classes it contains. Without them ROOT cannot apply schema evolution, so any change to an event class definition makes older files unreadable.

## Mechanism

1. Worker threads write events to temporary files; `TRestRun::MergeToOutputFile` merges them. The merged file at this point **has** all StreamerInfos.
2. The same method then reopens the output in UPDATE mode to write metadata.
3. On close, `TFile::WriteStreamerInfo()` (see ROOT `TFile.cxx`) frees the previous StreamerInfo record and writes only the classes streamed during the current session — i.e. only the metadata classes. The event-class infos are gone.

Verified with a minimal reproducer (write tree → infos present; TFileMerger → still present; reopen UPDATE + write a TNamed → event infos wiped). restG4 files are unaffected (single write session).

## Consequences
- rest-for-physics/detectorlib#125: the `vector<Float_t>` → `vector<Double_t>`  change in TRestDetectorSignal (detectorlib#109) made all pre-v2.4.3 detector  signal data unreadable (multi-GB allocations / OOM), because ROOT has no record of the on-disk layout. Recovery tool in #566. 
- Any future event-class change will have the same effect on files written before the fix.

## Fix

PR #567 re-tags the existing infos after each UPDATE open so they are rewritten on close. Not covered there (lower risk, could be follow-ups): `TFileMerger`'s own UPDATE path in `MergeToOutputFile(outputfilename != "")`, `TRestAnalysisTree::AddChainFile`, and the plot tools' UPDATE sessions (`TRestMetadataPlot`, `TRestDataSetPlot`). Read-only UPDATE sessions are safe (`WriteStreamerInfo` is a no-op unless something was streamed).

cmargalejo:
cmargalejo Large: 2025 [](https://github.com/rest-for-physics/framework/commits/cris_preserve_streamerinfos) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=rest-for-physics&utm_campaign=badge_info)  

Fixes #568

## *UPDATE* 2026-08-19:

Also updating this PR significantly. The idea was essentially correct, but it didn't go far enough. The issue is that with the PR in the original form there would be many edge cases where we could have lost the `StreamerInfo` in the future. Either through existing code paths or more importantly due to the way the developers of REST need to use the ROOT I/O in order to not lose that information.

- It centralizes writable ROOT access in a move-only `TRestRootFileHandle`.
- Before entering UPDATE mode, REST inventories and prepares the file’s historical `StreamerInfo` and schema rules, including classes whose dictionaries are not currently loaded.
- Existing direct writable `TFile` call sites are migrated to this checked path.
- ROOT-file merges become transactional: build and validate a temporary sibling, then replace with rollback support.
- Remote ROOT files remain valid inputs, but writable destinations must be local.
- `TRestRun` becomes non-copyable because it now owns file handles and contains non-owning aliases into their contents.
- Comprehensive tests cover loaded/unloaded dictionaries, automatic `vector<float> → vector<double>` evolution, explicit renamed-member rules, update failures, merges, rollback, and remote paths.

So #566 repairs the specific damaged legacy detector-signal files; #567 prevents normal REST updates and merges from discarding the historical schema information needed to read old data safely.

**Important note for developers**: In the future it is mandatory that all ROOT I/O happens through the new safe interop. See the documentation about this here:

https://github.com/rest-for-physics/framework/blob/07c720c6cd89f594c2047a3bbdff609c2e7f3c89/doc/developer/Safe%20writable%20ROOT%20IO.md

In principle this should make our lives *easier*.

The addition of the new I/O interface and updating of all existing ROOT file I/O code is the reason the PR exploded in size now. 

## The problem

  When a ROOT file is closed, `TFile::WriteStreamerInfo()` replaces the file's
  StreamerInfo record, keeping only the infos of the classes streamed during that
  session. `TRestRun::MergeToOutputFile` merges the threads' output files (which
  preserves the infos), but then reopens the merged file in UPDATE mode to write
  the metadata: closing that session wipes the StreamerInfos of all event classes,
  keeping only the metadata classes.

  As a consequence, **no restManager output file contains event-class
  StreamerInfos**, which silently breaks ROOT schema evolution whenever an event
  class definition changes. This is what made files written before the
  `TRestDetectorSignal` `vector<Float_t>` → `vector<Double_t>` change unreadable
  (rest-for-physics/detectorlib#125): without the on-disk layout description,
  ROOT misreads the float payload as doubles and allocates GBs of garbage.

  (restG4 files are unaffected: they are written in a single session.)

  ## The fix

  New `TRestTools::PreserveStreamerInfos(TFile*)`: called right after opening a
  file in UPDATE mode, it re-tags every StreamerInfo already stored in the file
  (same marking as the deprecated `TStreamerInfo::TagFile`) so that
  `TFile::WriteStreamerInfo` writes them out again on close. Called at the UPDATE
  sessions that write into REST data files:

  - `TRestRun::MergeToOutputFile` (the main restManager output path)
  - `TRestRun::UpdateOutputFile`
  - `TRestProcessRunner` split-file metadata update of the main file
  - `TRestDataSet::Export`

  ## Verification

  Merging two files containing `TRestDetectorSignalEvent` trees via
  `TRestRun::MergeToOutputFile`: without the fix the merged file keeps only
  `TRestRun`/`TRestMetadata` infos; with it, `TRestDetectorSignalEvent`,
  `TRestEvent` and `TRestDetectorSignal` survive the metadata session and the
  file reads back correctly.

  Note: this protects files written from now on. Existing files have already
  lost their StreamerInfos; for the detector signal case the data is recoverable
  with the tool in #566.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,6 +22,12 @@ It is recommended that new contributions and functionalities added to REST have 
 
 TODO : Explain doxygen formatting, tutorials, where official doc is located. ETC.
 
+### ROOT file I/O changes
+
+Code that creates, updates, replaces, or merges ROOT files must follow the
+[safe writable ROOT I/O guide](doc/developer/Safe%20writable%20ROOT%20IO.md). It documents the required
+schema preflight, ownership, transactional replacement, and local/remote path rules.
+
 ### Pipeline validation tests
 
 TODO : Explain how pipeline validation tests should be implemented
--- a/doc/developer/Safe writable ROOT IO.md
+++ b/doc/developer/Safe writable ROOT IO.md
@@ -1,0 +1,118 @@
+# Safe writable ROOT I/O
+
+REST ROOT files may contain several historical `TStreamerInfo` entries and embedded schema rules. Opening such
+a file directly with `TFile::Open(..., "UPDATE")` bypasses REST's schema preflight and can let ROOT rewrite
+schema metadata before REST has established that every historical class layout is usable. Framework code that
+creates or mutates ROOT files must therefore use `TRestRootFileHandle`.
+
+## Opening files
+
+Use `TRestRootFileHandle::Open` for new code:
+
+```cpp
+#include "TRestTools.h"
+
+auto file = TRestRootFileHandle::Open(filename, TRestRootFileMode::Update);
+if (!file) {
+    ReportError(file.Error());
+    return false;
+}
+
+file->cd();
+WriteObjects();
+
+if (!file.Close()) {
+    ReportError(file.Error());
+    return false;
+}
+```
+
+The available modes are `Read`, `Recreate`, and `Update`. `Recreate` intentionally replaces an existing file
+and must not be used as a shortcut for `Update`. An update is initially opened read-only. REST inventories the
+exact class-name, class-version, and checksum tuples stored in the file, collects the embedded schema rules,
+prepares loaded or emulated ROOT classes, and preflights the historical entries. It then transitions the same
+`TFile` to update mode and verifies the local file identity and ROOT UUID before marking the exact historical
+class-index entries required for writing.
+
+`PrepareBorrowedUpdate(TFile&, std::string*)` provides the same update preparation when legacy code already
+owns a `TFile`. The supplied file must be valid, open in `READ` mode, and not writable:
+
+```cpp
+std::unique_ptr<TFile> file(TFile::Open(filename.c_str(), "READ"));
+std::string error;
+if (!file || !TRestRootFileHandle::PrepareBorrowedUpdate(*file, &error)) {
+    ReportError(error);
+    return false;
+}
+```
+
+Prefer `TRestRootFileHandle` whenever ownership can be changed. A borrowed file remains the caller's
+responsibility, including checking its close/write status. If preparation fails, propagate the error and do not
+attempt to write through that file.
+
+The preflight preserves the historical StreamerInfo entries and schema rules it found on disk; it does not make
+an incompatible class change or an incorrect schema rule valid. Class authors must still increment class
+versions as required, write correct evolution rules, and test representative old files both before and after a
+writable open.
+
+## Ownership and error handling
+
+`TRestRootFileHandle` is move-only. Pass it by reference while it remains owned by a component, or transfer it
+with `std::move`. Pointers returned by `Get()` and `operator->` are non-owning and must not be deleted or retained
+beyond the handle's lifetime.
+
+The destructor closes an open file, but cannot report failure. Code that writes must call `Close()` explicitly
+and handle a false result using `Error()`. Close a live destination before move-assigning another handle to it,
+because move assignment cannot return a close error for the previous file.
+
+## Replacing or merging files
+
+Use `TRestTools::MergeRootFilesTransactionally` instead of merging directly into the destination or manually
+renaming a partially written file:
+
+```cpp
+std::string error;
+const std::string existing = outputAlreadyExists ? output : "";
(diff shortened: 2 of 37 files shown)


cmargalejo:

juanangp (approved):

Vindaar (approved):
The addition of the new changes *should* make it so that REST files should remain stable and readable in the future even under further schema changes. That's a big win for us, because there's nothing worse than breaking backwards compatibility for old data files. Especially given the context of the likely changes still to come for BabyIAXO once we get closer to actual data taking. At some point development will speed up and we are going to be happy this is done now and not later.

cmargalejo:
Hey @Vindaar, reading doc/developer/Safe writable ROOT IO.md, two small thoughts.

1. Since the doc points new code at the handle, I'm assuming existing code is fine and doesn't need a updating, right?. If it's useful, I'm happy to grep the framework + lib submodules for any leftover "UPDATE" opens, just to confirm nothing dangerous is still there.

2. For discoverability I'd lean on CI rather than the doc: a tiny check that fails on a direct TFile::Open(…, "UPDATE") (pointing to the handle) would catch it automatically, since people writing the occasional process won't read a doc/developer/ page anyway. 

What do you think?

Vindaar:
I've added a few more commits which mostly simplify the PR a little bit. The main points remain unchanged though.

Given the large size of the PR, here is an HTML page to review the code more easily, where we have a clearer overview of which files contain what kind of change, what to focus on when reviewing etc. You might find it useful:

https://pages.vindaar.de/rest-for-physics/pr567_compact_review.html



> Hey @Vindaar, reading doc/developer/Safe writable ROOT IO.md, two small thoughts.
> 
> 1. Since the doc points new code at the handle, I'm assuming existing code is fine and doesn't need a updating, right?. If it's useful, I'm happy to grep the framework + lib submodules for any leftover "UPDATE" opens, just to confirm nothing dangerous is still there.

Yes, there seem to be ~4 existing usages in detectorlib and tracklib that we still have to update. 

> 2. For discoverability I'd lean on CI rather than the doc: a tiny check that fails on a direct TFile::Open(…, "UPDATE") (pointing to the handle) would catch it automatically, since people writing the occasional process won't read a doc/developer/ page anyway.
> 
> What do you think?

Yes, this is a good idea in principle. We can have a small script that checks for existing lines that use the old style model and then write a warning or even CI failure if they are detected. The thing is, lines using `TFile::Open(..., "UPDATE")` are not guaranteed to break the `StreamerInfo` data, but there is a chance depending on what one does. But obviously things still continue to work, so we don't necessarily force every script to use it immediately. It's just preferable.

Vindaar:
> Given the large size of the PR, here is an HTML page to review the code more easily, where we have a clearer overview of which files contain what kind of change, what to focus on when reviewing etc. You might find it useful:
> 
> https://pages.vindaar.de/rest-for-physics/pr567_compact_review.html

Note, if you intend to review this PR, I recommend doing the following:
- read the main text on this page to understand the ideas of the changes better
- look at the list of files with changes, their categories and importance. You'll realize that the majority of changes are in test files (new tests, some changes to existing ones). 
- use the filter feature to filter only to files changed with impact `high`. Then you realize that the actual meat of this PR isn't all that big. In terms of lines of code it really only is additions to `src/TRestTools.cxx` which grew the new interface to handle ROOT I/O for us. 
- for each file (of the high impact changes), make sure to read the explanation at the top of that files's diff.

I think that makes the review a lot more manageable.

